PR #22, opened 2018-10-08
INFRA:332 Dockerize Wallet Daemon

thomsh:
Using openjdk 8 (u181) should be change/update via the Dockerfile
Using embeded libledger-core jar

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+apt-get update
+apt-get install -qy apt-transport-https software-properties-common
+apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 2EE0EA64E40A89B84B2DF73499E82A75642AC823
+add-apt-repository "deb https://dl.bintray.com/sbt/debian /"
+apt-get update && apt-get install -qy sbt
+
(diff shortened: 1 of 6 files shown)


docker/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+apt-get update
+apt-get install -qy apt-transport-https software-properties-common
+apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 2EE0EA64E40A89B84B2DF73499E82A75642AC823
+add-apt-repository "deb https://dl.bintray.com/sbt/debian /"
+apt-get update && apt-get install -qy sbt
```
rottenbytes: No fixed version of scala ?

thomsh: nope, it's define here : https://github.com/LedgerHQ/ledger-wallet-daemon/blob/f03d5d8aa22c332c9b7c9e5fd64bdb6d2c710ccd/build.sbt#L6

rottenbytes: I mean the package version here. how does it align with the aforementioned version ?

thomsh: We can fix it with `apt-get install sbt=1.2.3`.
But like python pip we use the latest version and  fix python version + lib version in requirements.txt or Pipfile.lock . So it's the same here with scala

PR #27, opened 2018-10-22
TEST CIRCLECI INFRA-344: Refresh Docker image through CircleCI

dud225:
This is a preliminary work to implement automatic rebuild of the Docker
image through CircleCI.


dud225:
It is a test PR to check that the CircleCI configuration behaves as expected.

PR #28, opened 2018-10-22
Add endpoint to validate address string

tt810:
1. Add endpoint for address string validation
2. Remove/update failed tests


daemon/src/test/scala/co/ledger/wallet/daemon/api/CurrenciesApiTest.scala
```diff
@@ -27,7 +34,28 @@ class CurrenciesApiTest extends APIFeatureTest {
   test("CurrenciesApi#Get currencies returns all") {
     val response: Response = assertCurrencies(CURRENCY_POOL, Status.Ok)
     val currencies: List[CurrencyView] = parse[List[CurrencyView]](response)
-    assert(currencies == List(EXPECTED_BTC_TESTNET, EXPECTED_BTC_CURRENCY))
+    assert(currencies.map(_.name) == List(
```
pollastri-pierre: This assert is a bit long. Maybe we can just test that currencies contains at list all of these. It's not the goal of the daemon to keep this list up to date.

daemon/src/test/scala/co/ledger/wallet/daemon/database/DaemonCacheTest.scala
```diff
@@ -60,10 +60,9 @@ class DaemonCacheTest extends AssertionsForJUnit {
 
   @Test def verifyGetCurrencies(): Unit = {
     val currencies = Await.result(cache.getCurrencies("pool_1", PUB_KEY_1), Duration.Inf)
-    assertEquals(2, currencies.size)
+    assertEquals(21, currencies.size)
```
pollastri-pierre: Same thing as previously should something like assert(currencies.size >= 21)

PR #29, opened 2018-10-23
Remove caching from Wallet class

tt810:
Accounts in wallet will not be cached. Now, the account are ordered by their creation order.


daemon/src/main/scala/co/ledger/wallet/daemon/models/Wallet.scala
```diff
@@ -153,26 +161,20 @@ class Wallet(private val coreW: core.Wallet, private val pool: Pool) extends Log
...
-        } else {
-          coreW.getAccounts(offset, count).map { coreAs =>
-            coreAs.asScala.foreach { coreA =>
-              cachedAccounts.put(coreA.getIndex, Account.newInstance(coreA, self))
-              debug(s"Add ${cachedAccounts(coreA.getIndex)} to $self cache")
-              cachedAccounts(coreA.getIndex).startRealTimeObserver()
-            }
-            accountLen.updateAndGet(n => Math.max(n, count))
-          }
+  private def startListen: Future[Unit] = {
```
pollastri-pierre: Side effects function should have parenthesis (to keep the code more expressive). Keep function without parenthesis for pure functions.

PR #31, opened 2018-10-24
log correlation ID send from gate

drakoFukayu:
So we can follow CID in all services, i.e. gate, hsm and daemon

### DIFF
--- a/daemon/src/main/scala/co/ledger/wallet/daemon/filters/CorrelationIdFilter.scala
+++ b/daemon/src/main/scala/co/ledger/wallet/daemon/filters/CorrelationIdFilter.scala
@@ -1,0 +1,14 @@
+package co.ledger.wallet.daemon.filters
+
+import com.twitter.finagle.{Service, SimpleFilter}
+import com.twitter.finagle.http.Request
+import org.slf4j.MDC
+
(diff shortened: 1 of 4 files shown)


daemon/src/main/scala/co/ledger/wallet/daemon/filters/CorrelationIdFilter.scala
```diff
@@ -0,0 +1,14 @@
+package co.ledger.wallet.daemon.filters
+
+import com.twitter.finagle.{Service, SimpleFilter}
+import com.twitter.finagle.http.Request
+import org.slf4j.MDC
+
```
tt810: could you add the comment header to your intellij and generate a doc:
`/**
  * Describe your class here.
  *
  * User: Ting Tu
  * Date: ${DAY}-${MONTH}-${YEAR}
  * Time: ${TIME}
  *
  */`

daemon/src/main/scala/co/ledger/wallet/daemon/filters/CorrelationIdFilter.scala
```diff
@@ -0,0 +1,23 @@
...
+/**
+  *
```
tt810: ok, you are not allowed to commit on scala code anymore

PR #34, opened 2018-10-24
Add summary of operation count in account view

tt810:


KhalilBellakrid:
It seems good to me !

PR #45, opened 2018-11-08
Remove Operation class

tt810:
Remove Operation class from Daemon, reuse the object from core-lib.


daemon/src/main/scala/co/ledger/wallet/daemon/database/DefaultDaemonCache.scala
```diff
@@ -188,51 +191,41 @@ class DefaultDaemonCache() extends DaemonCache with Logging {
...
+      previous = candidate.previous
+      operationRecord = opsCache.insertOperation(candidate.id, pool.id, walletName, accountIndex, candidate.offset(), candidate.batch, next, previous)
+      opsView <- Future.sequence(ops.map { op => Operations.getView(op, wallet, account)})
+    } yield PackedOperationsView(operationRecord.previous, operationRecord.next, opsView)
+  }
+
+  def getHardAccount(user: User, poolName: String, walletName: String, accountIndex: Int): Future[(Pool, Wallet, Account)] = {
+    for {
+      poolOpt <- user.pool(poolName)
+      pool = poolOpt.getOrElse(throw exceptions.WalletPoolNotFoundException(poolName))
```
luchenyuxx: Use `Future.failed` instead of `throw`, it's more efficient

daemon/src/main/scala/co/ledger/wallet/daemon/database/DefaultDaemonCache.scala
```diff
@@ -188,51 +191,41 @@ class DefaultDaemonCache() extends DaemonCache with Logging {
...
+      walletOpt <- pool.wallet(walletName)
+      wallet = walletOpt.getOrElse(throw exceptions.WalletNotFoundException(walletName))
```
luchenyuxx: Use `Future.failed` instead of `throw`, it's more efficient

daemon/src/main/scala/co/ledger/wallet/daemon/database/DefaultDaemonCache.scala
```diff
@@ -188,51 +191,41 @@ class DefaultDaemonCache() extends DaemonCache with Logging {
...
+      accountOpt <- wallet.account(accountIndex)
+      account = accountOpt.getOrElse(throw exceptions.AccountNotFoundException(accountIndex))
```
luchenyuxx: Use `Future.failed` instead of `throw`, it's more efficient

daemon/src/main/scala/co/ledger/wallet/daemon/services/AccountsService.scala
```diff
@@ -74,10 +77,15 @@ class AccountsService @Inject()(defaultDaemonCache: DaemonCache) extends DaemonS
   }
 
   def accountOperation(user: User, uid: String, accountIndex: Int, poolName: String, walletName: String, fullOp: Int): Future[Option[OperationView]] = {
-    defaultDaemonCache.getAccountOperation(user, uid, accountIndex, poolName, walletName, fullOp).flatMap {
-      case Some(op) => op.operationView.map(Option(_))
-      case None => Future.successful(None)
-    }
+    (for {
+      walletOpt <- defaultDaemonCache.getWallet(walletName, poolName, user.pubKey)
+      wallet = walletOpt.getOrElse(throw new WalletNotFoundException(s"Wallet '$walletName' not found"))
```
luchenyuxx: Use `Future.failed` instead of `throw`, it's more efficient

daemon/src/main/scala/co/ledger/wallet/daemon/services/AccountsService.scala
```diff
@@ -74,10 +77,15 @@ class AccountsService @Inject()(defaultDaemonCache: DaemonCache) extends DaemonS
...
+      accountOpt <- wallet.account(accountIndex)
+      account = accountOpt.getOrElse(throw new AccountNotFoundException(s"Account '$accountIndex' not found"))
```
luchenyuxx: Use `Future.failed` instead of `throw`, it's more efficient

daemon/src/main/scala/co/ledger/wallet/daemon/models/Operations.scala
```diff
@@ -0,0 +1,102 @@
...
+      indicator.getTrustLevel,
+      indicator.getConflictingOperationUids.asScala.toList,
+      indicator.getOrigin)
+  }
+
+  def getTransactionView(operation: core.Operation, curFamily: core.WalletType): Option[TransactionView] = {
+    if (operation.isComplete) {
+      curFamily match {
+        case core.WalletType.BITCOIN => Some(Bitcoin.newTransactionView(operation.asBitcoinLikeOperation().getTransaction))
+        case _ => throw new UnsupportedOperationException
```
luchenyuxx: You don't need to match this case, since it won't happen. Scala warns you for missing cases. In the future, if new cases are added, you are forced to solve those cases here. If you match `_`, new cases won't be prompted here.

daemon/src/main/scala/co/ledger/wallet/daemon/models/Account.scala
```diff
@@ -85,25 +85,22 @@ object Account {
         } else throw new UnsupportedOperationException("Account type not supported, can't create transaction")
     }
 
-    def operation(uid: String, fullOp: Int): Future[Option[Operation]] = {
+    def operation(uid: String, fullOp: Int): Future[Option[core.Operation]] = {
```
luchenyuxx: `fullOp` should be `boolean` if it's only used as a flag

daemon/src/main/scala/co/ledger/wallet/daemon/database/DefaultDaemonCache.scala
aymeric-ledger: This actually throws here.

We can write it with `Future.failed` and `Future.successful`, however it's not very elegant. A solution would be to return a `Future[Pool]` with `user.pool`, or a `Future[Option[(Pool, Wallet, Account)]]` with `getHardAccount`, whichever makes the most sense.

luchenyuxx: We rely on the exception to handle response. So `option` is not suit here.
You can consider `Future` as an `Either[Exception, Data]`, maybe this makes more sense for you to use `Future.failed`

aymeric-ledger: `Future.failed` is fine by me, but couldn't we return a `Future[Pool]` directly from `user.pool` ?

Otherwise it will looks something like this:
`pool <- poolOpt.fold[Future[Pool]](Future.failed(exceptions.WalletPoolNotFoundException(poolName)))(Future.successful)`

luchenyuxx: oh, I understand your point. I think it depends on when do we want to handle the `None` case. If we return `Future[Pool]`, the logic expecting a `Option[Pool]` will have to handle the exception. I admit the concept is a little mixed here. I am OK as along as we don't use `throw`.

daemon/src/main/scala/co/ledger/wallet/daemon/models/Account.scala
tt810: I don't want to make it a boolean which will change the type in the future if there are more options from lib-core which need to be handled by Gate.

PR #57, opened 2018-12-03
LV-803: Add outbound proxy support

jchampemont:


daemon/src/main/scala/co/ledger/wallet/daemon/clients/ApiClient.scala
```diff
@@ -26,9 +28,17 @@ import scala.concurrent.{ExecutionContext, Future}
 class ApiClient(implicit val ec: ExecutionContext) {
   import ApiClient._
   private[this] val (host, port, poolSize) = DaemonConfiguration.apiConnection
-  private [this] val client = Http.client
-    .withSessionPool.maxSize(poolSize)
-    .newService(s"$host:$port")
+  private[this] val (proxyEnabled, proxyHost, proxyPort) = DaemonConfiguration.proxy
```
aymeric-ledger: I doubt the package private is meaningful here. Is it ?

jchampemont: Not really, I used the same visibility as the other stuff in here.

PR #58, opened 2018-12-03
Use environment variables for configuration

jchampemont:


jchampemont:
Infra asked me to put environment variables in the config file in order for them to be able to spin up a docker container using environment variables only. This is already done in the driver (see https://github.com/LedgerHQ/ledger-vault-hsm-driver/blob/develop/samples/application.prod.conf#L146 for instance)

PR #67, opened 2018-12-31
ETH ERC20 support, refactoring.

luchenyuxx:
This PR contains big changes:

## Project structure change. 

The sub project `lib-core-binding` is removed. The whole project now have no sub project. The reason is `lib-core-binding` contains only the java and scala codes generated by `lib-core`'s interface description. They are bound to `lib-core`'s c++ implementation and not supposed to be manipulated by daemon developer.

Instead, the Java and Scala sources in `lib-core-binding` are now compiled and packaged together with the `lib-core` native library in `ledger-lib-core.jar`. So that, the Java and Scala interface is always matching the c++ implementation.

A helper script is added to build the `ledger-lib-core.jar`.

Some CI related script is updated to fix CI failure due to the change of the project structure.

## ETH and ERC20 support

ETH related endpoint is integrated with BTC endpoint. The user should have no change on their query endpoint, just need to parse the returned ETH response body.

To create ETH account, only one derivation path is required.

ERC20 account is actually a ETH account. It's identified by a token address. ERC20 has their own endpoints.

Currently, we don't have pagination on ERC20 operation query.

For details, see LV-735

## Refactoring

Some methods in `DaemonCache` are removed, since they are only used in one place, and clearly not a cache related function. This simplified the `DaemonCache`.

Added new configs to support assigning different explorer URLs for different coins. Since BTC and ETH use different explorers.

Refactored the Account Controller code to tree like style to avoid boilerplates.

## Related tickets

- LV-735: [DAEMON] Include ETHEREUM
- LV-849: WALLET DAEMON: remove subproject `ledger-core-binding`
- LV-850: WALLET DAEMON: refactor controllers
- LV-852: WALLET DAEMON: Add type to represent account path
- LV-851: WALLET DAEMON: refactor DaemonCache
- LV-854: [DAEMON]: remove toString override methods for case classes
- LV-887: [Wallet Daemon] remove unnecessary methods in DaemonCache
- LV-888: [Wallet Daemon] Remove build lib core process in CI building after LV-849


build-jar.sh
```diff
@@ -0,0 +1,118 @@
...
+  sbt package
+  cd $CURRENT_DIR
+}
+
+# cleanup
+function cleanup()
+{
+  clean_src
+  docker rm -f $DOCKER_NAME
+}
```
jchampemont: I would add `trap cleanup EXIT` that will always run this function on exit.

docker/build_ledger_core.sh
```diff
@@ -3,7 +3,9 @@ set -euxo pipefail
 
 # Clone the lib-core at the target commit
 git clone --recurse-submodules https://github.com/LedgerHQ/lib-ledger-core.git
-(cd lib-ledger-core && git reset --hard "$(head -n 1 ../ledger_core_commit)")
+(cd lib-ledger-core && git remote add khalil https://github.com/KhalilBellakrid/lib-ledger-core.git && git fetch khalil)
```
jchampemont: Why are we using Khalil's fork here?

package_libcore.sh
```diff
@@ -25,15 +25,18 @@ fi
...
 cd ${current_dir}
 rm -rf -- ${pkg_dir}
 echo "Libraries updated"
 
 echo "Updating APIs"
-rm -f -- ${java_api_dir}/*
-cp ${libcore_dir}/api/core/java/* ${java_api_dir}
-rm -f -- ${scala_api_dir}/implicits/*
-cp ${libcore_dir}/api/core/scala/* ${scala_api_dir}/implicits/
+# TODO add back copies once all is fixed
```
jchampemont: I would open a Jira and reference the Jira ticket number here to make sure we do not forget this TODO.

src/main/scala/co/ledger/wallet/daemon/database/DaemonCache.scala
```diff
@@ -0,0 +1,266 @@
...
+
+
+  /**
+    * Getter of account instance with specified parameters.
+    *
+    * @param accountIndex the unique index of specified instance.
+    * @param pubKey       the public key of instance of `co.ledger.wallet.daemon.DefaultDaemonCache.User`.
+    * @param poolName     the name of wallet pool the account belongs to.
+    * @param walletName   the name of wallet the account belongs to.
+    * @return a Future of an Option of the instance of `co.ledger.wallet.daemon.models.Account`.
```
jchampemont: The scaladoc is not up to date with the function signature. (same thing for a dozen other functions in this file).

src/main/scala/co/ledger/wallet/daemon/models/package.scala
```diff
@@ -0,0 +1,16 @@
+package co.ledger.wallet.daemon
+
+/**
+  * Describe your class here.
```
jchampemont: Replace default scaladoc.

src/main/scala/co/ledger/wallet/daemon/services/AccountsService.scala
```diff
@@ -0,0 +1,179 @@
...
+        ao.map(_.accountView(accountInfo.walletName, wallet.getCurrency.currencyView)).sequence)
+    }
+  }
+
+  /**
+    * Method to synchronize account operations from public resources. The method may take a while
+    * to finish. This method onlly synchronize a single account.
+    *
+    * @return a Future of sequence of result of synchronization.
+    */
```
jchampemont: The scaladoc is not up to date with the function signature. (same thing for a couple other functions in this file).

src/test/scala/co/ledger/wallet/daemon/api/CurrenciesApiTest.scala
```diff
@@ -67,7 +67,7 @@ class CurrenciesApiTest extends APIFeatureTest {
 
   private val CURRENCY_POOL = "currency_pool"
   private val CURRENCY_BTC = "bitcoin"
-  private val CURRENCY_NON_EXIST = "ethereum"
+  private val CURRENCY_NON_EXIST = "currency_non_exist"
```
jchampemont: 👍

docker/build_ledger_core.sh
luchenyuxx: deleted this file, since we don't need it anymore

build-jar.sh
luchenyuxx: 👍 learn a new thing

package_libcore.sh
luchenyuxx: deleted this file, since we don't need it with the new project structure

src/main/scala/co/ledger/wallet/daemon/database/DaemonCache.scala
luchenyuxx: Removed unnecessary comments

src/main/scala/co/ledger/wallet/daemon/models/package.scala
luchenyuxx: removed

src/main/scala/co/ledger/wallet/daemon/models/coins/Ethereum.scala
```diff
@@ -0,0 +1,136 @@
...
+  def apply(n: EthereumLikeNetworkParameters): EthereumNetworkParamView =
+    EthereumNetworkParamView(
+      n.getIdentifier,
+      n.getMessagePrefix,
+      HexUtils.valueOf(n.getXPUBVersion),
+      n.getAdditionalEIPs.asScala.toList,
+      n.getTimestampDelay)
+}
+
+case class EthereumBlockView(
```
tt810: This seems the same as BitcoinBlockView, is it better to use the same object ?

luchenyuxx: yep, going to change

PR #77, opened 2019-01-09
Lv 795

luchenyuxx:
add error codes to error response body


luchenyuxx:
Only the bad request response error code is checked. Since the not found response won't have the error code...

PR #78, opened 2019-01-11
Fetch the estimated gas limit if not provided

YBadiss:


src/main/scala/co/ledger/wallet/daemon/clients/ApiClient.scala
```diff
@@ -65,13 +65,23 @@ class ApiClient(implicit val ec: ExecutionContext) {
...
   }
 
-  def getGas(currencyName: String): Future[GasInfo] = {
+  def getGasLimit(currencyName: String, recipient: String): Future[BigInt] = {
+    val (host, service) = hostServices.getOrElse(currencyName, hostServices("default"))
+    val request = Request(Method.Get, s"/blockchain/v3/addresses/$recipient/estimate-gas-limit")
+    request.host = host
+
+    service(request).map { response =>
+      mapper.readValue(response.contentString, classOf[GasInfo]).limit
```
tt810: This is weird, I am not sure it can pass. the response here should just contain a single field, it won't be parsed to `GasInfo`

src/main/scala/co/ledger/wallet/daemon/clients/ApiClient.scala
```diff
@@ -65,13 +65,23 @@ class ApiClient(implicit val ec: ExecutionContext) {
...
+  }
+
+  def getGasPrice(currencyName: String): Future[BigInt] = {
     val (host, service) = hostServices.getOrElse(currencyName, hostServices("default"))
     val request = Request(Method.Get, "/blockchain/v3/fees")
     request.host = host
 
     service(request).map { response =>
-      mapper.readValue(response.contentString, classOf[GasInfo])
+      mapper.readValue(response.contentString, classOf[GasInfo]).price
```
tt810: the same here. You will probably need to readValue to another case class (something like GasPrice) to get the price.

YBadiss: It works in both cases, I ran a test. But yeah if it looks weird to you I'll create two separate types

PR #98, opened 2019-02-27
Increase the time of token validity

YBadiss:
This hotfix was done on master during the latest release and never applied on dev.
https://github.com/LedgerHQ/ledger-wallet-daemon/pull/95

It extends the validity time of the auth tokens sent by the gate. A more in-depth work is required to remove this auth as much as possible.

### DIFF
--- a/src/main/scala/co/ledger/wallet/daemon/configurations/DaemonConfiguration.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/configurations/DaemonConfiguration.scala
@@ -12,7 +12,7 @@ import scala.util.Try
 object DaemonConfiguration {
   private val config = ConfigFactory.load()
   private val PERMISSION_CREATE_USER: Int = 0x01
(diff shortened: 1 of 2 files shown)


src/main/scala/co/ledger/wallet/daemon/configurations/DaemonConfiguration.scala
```diff
@@ -12,7 +12,7 @@ import scala.util.Try
 object DaemonConfiguration {
   private val config = ConfigFactory.load()
   private val PERMISSION_CREATE_USER: Int = 0x01
-  private val DEFAULT_AUTH_TOKEN_DURATION: Int = 30000 // 30 seconds
+  private val DEFAULT_AUTH_TOKEN_DURATION: Int = 3600 * 1000 // 30 seconds
```
luchenyuxx: update also the comment

PR #100, opened 2019-03-05
Skip unsupported currencies such as XRP

YBadiss:
When building a `currencyView` for a currency not supported by the Daemon, we raise an exception. This prevents listing all currencies when one of them has not been yet implemented on the Daemon and by extension on the gate.

By skipping the currencies generating exceptions, we make ourselves more resilient to libcore changes.


luchenyuxx:
See #101

PR #101, opened 2019-03-05
return UnsupportedNetworkParamsView than throw an exception for unsup…

luchenyuxx:
…ported coin type

### DIFF
--- a/src/main/scala/co/ledger/wallet/daemon/models/Currency.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/models/Currency.scala
@@ -8,12 +8,18 @@ import com.fasterxml.jackson.annotation.JsonProperty
 import scala.collection.JavaConverters._
 
 object Currency {
+
   implicit class RichCoreCurrency(val c: core.Currency) extends AnyVal {
     def concatSig(sig: Array[Byte]): Array[Byte] = Currency.concatSig(c)(sig)
+
     def parseUnsignedBTCTransaction(rawTx: Array[Byte], currentHeight: Long): Either[String, core.BitcoinLikeTransaction] = Currency.parseUnsignedBTCTransaction(c)(rawTx, currentHeight)
+
     def parseUnsignedETHTransaction(rawTx: Array[Byte]): Either[String, core.EthereumLikeTransaction] = Currency.parseUnsignedETHTransaction(c)(rawTx)
+
     def validateAddress(address: String): Boolean = Currency.validateAddress(c)(address)
+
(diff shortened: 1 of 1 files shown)


luchenyuxx:
@YBadiss please check if this meet your need.

src/main/scala/co/ledger/wallet/daemon/models/Currency.scala
```diff
@@ -50,13 +56,25 @@ object Currency {
...
+    * Lib core gives all the coins it supports when asked from WalletPool,
+    * some of them may not be supported yet by wallet daemon. So, throwing
+    * an exception here prevents people to list even supported coin.
+    */
   private def newNetworkParamsView(coreCurrency: core.Currency): NetworkParamsView = coreCurrency.getWalletType match {
     case core.WalletType.BITCOIN => Bitcoin.newNetworkParamsView(coreCurrency.getBitcoinLikeNetworkParameters)
     case core.WalletType.ETHEREUM => EthereumNetworkParamView(coreCurrency.getEthereumLikeNetworkParameters)
-    case _ => throw new UnsupportedOperationException
+    case core.WalletType.RIPPLE =>
+      UnsupportedNetworkParamsView("RIPPLE", "wallet type RIPPLE is not supported by wallet daemon")
```
YBadiss: actually, wouldn't it be a bit cleaner if we used the `core.WalletType` value? also it would be more generic:

```scala
case otherWalletType => UnsupportedNetworkParamsView(otherWalletType)

//...
case class UnsupportedNetworkParamsView(walletType: core.WalletType) extends NetworkParamsView {
    def toString(): String{
        s"Wallet type $walletType is not supported by Wallet Daemon"
    }
}
```

or something like that?

luchenyuxx: Our JSON library read class parameter to produce json field. That's why we don't do `toString`

luchenyuxx:
@YBadiss I think we can live with this for the moment? I feel not right how ever I try to handle it. Or we downgrade the lib core

YBadiss:
Well as we saw together, removing the exception affects other parts of the code and could result in harder to track errors.

After rethinking it, it seems better to keep the exception and catch early on only in some cases.

If you're not happy with either solution let's quickly discuss it with @aymeric-ledger or @jchampemont tomorrow morning

YBadiss:
SHould we close this @luchenyuxx ?

luchenyuxx:
closed

PR #103, opened 2019-03-07
Retrieve explorer and ws endpoints from env var

YBadiss:


src/main/resources/application.conf.sample
```diff
@@ -81,25 +81,31 @@ explorer = {
       }
       {
         currency = ethereum
-        host = "http://eth-mainnet.explorers.prod.aws.ledger.fr"
+        host = "http://eth-mainnet.explorers.dev.aws.ledger.fr"
+        host = ${?WALLET_ETH_EXPLORER}
         port = 80
```
thomsh: If this variable is used ?  it should be good to add port at the same time (in env var)

src/main/resources/application.conf.sample
```diff
@@ -81,25 +81,31 @@ explorer = {
...
     ]
   }
   ws = {
     default = "wss://ws.ledgerwallet.com/blockchain/v2/{}/ws"
     ethereum = "ws://notification.explorers.dev.aws.ledger.fr:9000/ws/{}"
+    ethereum = ${?WALLET_ETH_WS_ENDPOINT}
     ethereum_classic = "ws://notification.explorers.dev.aws.ledger.fr:9000/ws/{}"
+    ethereum_classic = ${?WALLET_ETC_WS_ENDPOINT}
     ethereum_ropsten = "ws://notification.explorers.dev.aws.ledger.fr:9000/ws/{}"
+    ethereum_ropsten = ${?WALLET_ETH_ROPSTEN_ENDPOINT}
```
thomsh: name should be `WALLET_ETH_ROPSTEN_WS_ENDPOINT` no?

thomsh: Maybe add `NOTIFICATION` in the name ?

src/main/resources/application.conf.sample
```diff
@@ -81,25 +81,34 @@ explorer = {
...
-        host = "http://eth-ropsten.explorers.prod.aws.ledger.fr"
+        host = "http://eth-ropsten.explorers.dev.aws.ledger.fr"
+        host = ${?WALLET_ETH_ROPSTEN_EXPLORER_ENDPOINT}
         port = 80
+        port = ${?WALLET_ETH_ROPSTEN_EXPLORER_PORT}
       }
     ]
   }
   ws = {
     default = "wss://ws.ledgerwallet.com/blockchain/v2/{}/ws"
```
thomsh: Miss this one with an env var ? right?

PR #105, opened 2019-03-11
Release candidate

YBadiss:
Changelog:
- Increased gate token validity time
- Bump libcore to 2.7.0 (support for erc20 balances and added ripple)
- Fixed issue linked to the addition of ripple in libcore
- Updated configuration to allow overriding explorer and WS endpoints


YBadiss:
Actually let's do that next week with the rest of the services

PR #107, opened 2019-03-13
Sync configuration for all wallets on startup

aymeric-ledger:


YBadiss:
@luchenyuxx @aymeric-ledger please take a look, this is now tested with the latest version of the libcore

aymeric-ledger:
Looks fine to me, as far as the wallet daemon goes ! I would approve, but it's actually my own PR 😛

KhalilBellakrid:
@YBadiss could we please have some QA before merging this ? Vault should not use it before having this !

YBadiss:
@KhalilBellakrid this is only being merged on develop, once it is done we will deploy it our preprod env for further testing. Only after that will it be deployed on production.

Also note that I tested this fix locally.

PR #115, opened 2019-03-28
Feature/improve build

YBadiss:
The version endpoint now provides:
- commit hash of the wallet daemon
- version of the libcore
- explorer endpoints

Regarding the libcore, we would want to get the commit hash but this is not doable yet, I created https://ledgerhq.atlassian.net/browse/LLC-190 for the core team


build.sbt
```diff
@@ -1,4 +1,6 @@
-name := "daemon"
+name := "wallet-daemon"
+
+version := "2.0.0"
```
thomsh: If this will be always replace by a sed set a default value to "changeme" "undef" or "0.0.0" no ?

build.sbt
```diff
@@ -116,3 +118,11 @@ libraryDependencies ++= Seq(
   compilerPlugin("com.github.ghik" %% "silencer-plugin" % "1.3.1"),
   "com.github.ghik" %% "silencer-lib" % "1.3.1" % Provided
 )
+
+// For sbt plugin sbt-buildinfo
+lazy val root = (project in file(".")).
```
luchenyuxx:
```suggestion
enablePlugins(BuildInfoPlugin)
buildInfoKeys := Seq[BuildInfoKey](name, version, scalaVersion, sbtVersion)
buildInfoPackage := "co.ledger.wallet.daemon"
```

We are using flat style sbt file (no project definition), so do like this

build.sbt
```diff
@@ -1,4 +1,6 @@
-name := "daemon"
+name := "wallet-daemon"
+
+version := "unknown-commit-hash" // Set automatically on build
```
luchenyuxx: It's very weird to have the behaviour like this. If it's set by the CI, how could CI know the version?

YBadiss: the CI takes the current hash as the version, so whatever is being built is the version

build.sbt
YBadiss: This is the recommended way of using this plugin, see https://github.com/sbt/sbt-buildinfo#usage

If I remove this level, the compilation step fails because of BuildInfo

build.sbt
luchenyuxx: To me, the hash can be a build info, but it's weird to be a version number

YBadiss: it's not a version number, just the version

build.sbt
luchenyuxx: Tested on my machine, the compilation won't fail.

build.sbt
```diff
@@ -1,5 +1,9 @@
 name := "daemon"
 
+version := "0.0.0"
+val commitHash = SettingKey[String]("commitHash")
```
luchenyuxx: why not just `lazy val commitHash = sys.env.getOrElse("COMMIT_HASH", "unknown-commit-hash")`?

jchampemont: `sbt-git` should be used to retrieve git commit hashes. This is what we are using in crypto assets registry and countervalue API.

Please check this PR for integration details:
https://github.com/LedgerHQ/crypto-assets-registry/pull/42/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR72

YBadiss: updated build info see below

YBadiss: @jchampemont can't actually use this, as the build happens in the docker container, where we don't have the git repo anymore. So I have to extract the hash and pass it to the container

PR #117, opened 2019-04-03
Expose endpoint to retrieve account xpub

YBadiss:
We need to expose the pubkey to the frontend for clients who want to compute balances on their end.


src/main/scala/co/ledger/wallet/daemon/controllers/AccountsController.scala
```diff
@@ -280,6 +286,13 @@ object AccountsController {
                              request: Request
                            ) extends BaseSingleAccountRequest
 
+  case class XpubRequest(
```
luchenyuxx: It's the same as `AccountRequest`

luchenyuxx: Use `AccountRequest` instead

src/main/scala/co/ledger/wallet/daemon/controllers/AccountsController.scala
```diff
@@ -114,6 +114,12 @@ class AccountsController @Inject()(accountsService: AccountsService) extends Con
         accountsService.getBalance(request.contract, request.accountInfo)
       }
 
+      // End point queries for account xpub
+      get("/xpub") { request: XpubRequest =>
```
luchenyuxx: Don't have to add a new request type `XpubRequest`, use `AccountRequest` instead
```suggestion
      get("/xpub") { request: AccountRequest =>
```

PR #121, opened 2019-06-13
Handle new estimated gas limit format

YBadiss:
The old format of the gas limit request to the explorers was `{"estimated_gas_limit": 21000}` but we now receive `{"to": "0xeE076b1050A9f39671Cf54a7d2c9E305B0cb12c5", "gas_limit": 21000}`.

I believe this call is now integrated in the libcore but I don't have the time to look into it. So we keep our hack-ish direct call to the explorer and we now handle both cases as we are not sure yet which explorer is going to be running in production when we release.


luchenyuxx:
This is caused by the bug of explorer, Fixed in BACK-173.
Closing the PR.

PR #126, opened 2019-06-17
Send input and address to retrieve gas limit estimate

YBadiss:
The new version of the explorer provides a more accurate estimate for
the gas limit, but it now requires extra data for erc20 tokens.

I split the transaction creation to be more readable and added a default
gas limit of 200k to use when nothing is returned by the explorer.


src/main/scala/co/ledger/wallet/daemon/clients/ApiClient.scala
```diff
@@ -33,12 +36,21 @@ class ApiClient(implicit val ec: ExecutionContext) {
...
-  def getGasLimit(currencyName: String, recipient: String): Future[BigInt] = {
+  def getGasLimit(currencyName: String, recipient: String, source: Option[String] = None, inputData: Option[Array[Byte]] = None): Future[BigInt] = {
     val (host, service) = services.getOrElse(currencyName, services("default"))
-    val request = Request(Method.Get, s"/blockchain/v3/addresses/${recipient.toLowerCase}/estimate-gas-limit")
-      .host(host)
+    val request = Request(
+      Method.Get,
+      s"/blockchain/v3/addresses/${recipient.toLowerCase}/estimate-gas-limit"
+    ).host(host)
+    val body = Map("from" -> source.getOrElse(""), "data" -> s"0x${inputData.map(HexUtils.valueOf).getOrElse("")}")
```
luchenyuxx: `from` should be a correct address if set. Empty hex value won't pass the node.

PR #129, opened 2019-07-09
Fix/use latest libcore and handle erc20 data

YBadiss:
- Use latest libcore 3.0.0 (6214a98ceb273d0afc07038c14aa7f5328fad935) to have all fixes on erc20 balances etc.
- Update the erc20 balance usage to support Future
- Retrieve erc20 operation details to build full operation view without hacky parsing of eth data


src/main/scala/co/ledger/wallet/daemon/models/Account.scala
```diff
@@ -29,13 +29,13 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
 object Account extends Logging {
 
   implicit class RichCoreAccount(val a: core.Account) extends AnyVal {
-    def erc20Balance(contract: String): Either[Exception, scala.BigInt] =
+    def erc20Balance(contract: String)(implicit ec: ExecutionContext): Either[Exception, Future[scala.BigInt]] =
```
luchenyuxx: Not necessary to return `Either[Exception, Future[BigInt]]`, just `Future[BigInt]` is ok

PR #133, opened 2019-07-15
Master drop2

drakoFukayu:
oops some change where pushed directly to master-drop2
here is the new lib-core 3.0.0 and we have some debug message


drakoFukayu:
Was updating the libcore.
Indeed I did not pull the linux lib.
I push by mistake on master-drop2 was merging master in develop.

PR #136, opened 2019-07-16
Ripple integration

aymeric-ledger:
Integration made to mimic Ethereum integration.

Untested.


src/main/scala/co/ledger/wallet/daemon/controllers/TransactionsController.scala
```diff
@@ -136,11 +149,27 @@ object TransactionsController {
     def validateFees: ValidationResult = CommonMethodValidations.validateFees(feesPerByteValue, fees_level)
   }
 
+  case class XRPSendTo(amount: BigInt, address: String)
+
+  case class CreateXRPTransactionRequest(send_to: List[(String, String)], // Send funds to the given address.
```
KhalilBellakrid: I think you might need also `send_to` to be optional, otherwise in case of `wipe_to` mode ...

src/main/scala/co/ledger/wallet/daemon/models/Account.scala
```diff
@@ -65,6 +68,9 @@ object Account extends Logging {
     def broadcastETHTransaction(rawTx: Array[Byte], signatures: ETHSignature, c: core.Currency)(implicit ec: ExecutionContext): Future[String] =
       Account.broadcastETHTransaction(rawTx, signatures, a, c)
 
+    def broadcastXRPTransaction(rawTx: Array[Byte], signatures: ETHSignature, c: core.Currency): Future[String] =
```
KhalilBellakrid: You meant `XRPSignature` :trollface:

src/main/scala/co/ledger/wallet/daemon/models/coins/Ripple.scala
```diff
@@ -0,0 +1,82 @@
...
+    RippleNetworkParamView(
+      n.getIdentifier,
+      n.getMessagePrefix,
+      HexUtils.valueOf(n.getXPUBVersion),
+      n.getAdditionalRIPs.asScala.toList,
+      n.getTimestampDelay
+    )
+}
+
+case class RippleTransactionView(@JsonProperty("hash") hash: String,
```
KhalilBellakrid: You might need to add both optional fields `memos` and `destinationTag` to signed and unsigned views 🤔 no ?

src/test/resources/application.conf
```diff
@@ -52,17 +52,32 @@ core_data_path = "core_data_path"
...
+    connection_pool_size = 50
+    paths = [
+      {
+        currency = default
+        host = "api.ledgerwallet.com"
+        port = 80
+      }
+      {
+        currency = ethereum
+        host = "http://eth-mainnet.explorers.dev.aws.ledger.fr"
```
KhalilBellakrid: In case you need equivalent for XRP, you can use : `https://s2.ripple.com` for the moment, until we have our own node

build.sbt
```diff
@@ -3,10 +3,17 @@ name := "daemon"
 version := "0.0.0"
 
 organization := "co.ledger"
-scalaVersion := "2.12.2"
+scalaVersion := "2.12.7"
```
luchenyuxx: upgrade to `2.12.8`

src/main/scala/co/ledger/wallet/daemon/controllers/TransactionsController.scala
aymeric-ledger: It's a list that can be empty. I need to check how jackson handles this case, if it requires the field to be present or not.

src/main/scala/co/ledger/wallet/daemon/models/coins/Ripple.scala
aymeric-ledger: It made me realize I can get an array of memo, so I guess at the request level I need to use a list of memos? cc @KhalilBellakrid

PR #137, opened 2019-07-19
Merge pull request #135 from LedgerHQ/master-drop2

drakoFukayu:
Master drop2


KhalilBellakrid:
We should definitely get rid of those banches and come back to a normal flow !

PR #139, opened 2019-07-19
Clean up tests and CI

aymeric-ledger:


src/test/scala/co/ledger/wallet/daemon/clients/ApiClientTest.scala
```diff
@@ -6,14 +6,6 @@ import org.scalatest.junit.AssertionsForJUnit
 import scala.concurrent.Await
 import scala.concurrent.duration.Duration
 
-/**
-  * Test cases for [[ApiClient]].
-  *
-  * User: Ting Tu
```
YBadiss: :'(

src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,15 @@
+<configuration>
```
luchenyuxx: Why it's renamed to `logback-test`?

project/plugins.sbt
```diff
@@ -1,2 +1,6 @@
-addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager" % "1.3.25")
+addCompilerPlugin("org.psywerx.hairyfotr" %% "linter" % "0.1.17")
```
luchenyuxx: The author suggests putting this in `build.sbt`

src/test/resources/logback-test.xml
aymeric-ledger: I understand this is the standard way of doing it.

PR #142, opened 2019-08-08
Use the operation type of the erc20 op

YBadiss:
The type of the transaction (SEND, RECEIVE, NONE) needs to come from the erc20 transaction, not the eth one as it can be None for eth and RECEIVE for erc20.


YBadiss:
Recreating, not from my fork, to access the circle ci secrets

PR #144, opened 2019-08-14
Disable wallet update by default and wait for it to complete

YBadiss:
The update of wallet configuration is extra work that does not need to run on every restart of the daemon. Instead we now define conf variable that can be overridden to force an update.

We now also wait for the update of wallets to complete before finishing the startup warmup.


src/main/scala/co/ledger/wallet/daemon/modules/DaemonCacheModule.scala
```diff
@@ -47,6 +48,18 @@ object DaemonCacheModule extends TwitterModule {
       }
     }
 
+    def startSynchronization(): Unit = {
+      val scheduler = new ScheduledThreadPoolTimer(
```
luchenyuxx: I would prefer a singleton scheduler

luchenyuxx:
If the update is not a heavy work, it doesn't bother to do it everytime

src/main/scala/co/ledger/wallet/daemon/modules/DaemonCacheModule.scala
YBadiss: sounds better but I am trying to do as little change as possible so as to not impact prod before going on holiday. So if it's okay with you I'll keep it as is, which is how it was done so far

YBadiss:
> If the update is not a heavy work, it doesn't bother to do it everytime

@luchenyuxx true, but I am also removing it from the default run as I believe this update is somewhat linked to a recurring issue with transactions not being picked up by the wallet-daemon. 

Since endpoint updates are very infrequent we can enable this update feature when required.
If the issue I mentioned disappears from prod, then at least I'll know it was coming from here.

src/main/scala/co/ledger/wallet/daemon/modules/DaemonCacheModule.scala
luchenyuxx: alright, then please put a TODO there to not get forgotten.

PR #153, opened 2019-10-02
BACK-300: add session timeout

luchenyuxx:


src/main/scala/co/ledger/wallet/daemon/Server.scala
```diff
@@ -39,6 +41,12 @@ class ServerImpl extends HttpServer {
       .exceptionMapper[DaemonExceptionMapper]
       .exceptionMapper[LibCoreExceptionMapper]
 
+  override protected def configureHttpServer(server: Http.Server): Http.Server = {
+    server
+      .withSession.maxIdleTime(Duration.fromSeconds(10))
```
YBadiss: how does maxIdleTime work? some transaction queries take a very long time, are they going to be impacted?

PR #154, opened 2019-10-03
BACK-303: add status field in the ethereum transaction

luchenyuxx:


YBadiss:
do you know how should the status be interpreted? If so could the wallet-daemon transform the int into an enum that is easier to consume?

PR #169, opened 2019-11-07
Back-317 : Make WD tests great again

RomanWlm:
**This PR reenable tests on CI**

- All tests are now ITs => TODO : cleaning of tests (dead code) and separation of Tests/IT
- tests data are now generated under target (target/core_data_path) => Sbt clean to remove them, to be discussed but we could either generated a dedicated test directory for each runs or clean after suite theses data
- Fix Exception Handler was not triggered sinc Finatra is not able to deal with trait
- Whitelist test user to bypass authent properly
- ITs are using h2 for WD db-cache


aymeric-ledger:
Did not dig into this, but shouldn't we have some unit test coverage instead of only integration tests? Especially since the stack is not the most robust one.

RomanWlm:
> Did not dig into this, but shouldn't we have some unit test coverage instead of only integration tests? Especially since the stack is not the most robust one.

Absolutely ! There is some unit tests on it directory. This will be part of an other PR, test cleanings and segregation (cf PR description)

PR #170, opened 2019-11-08
Back 350 deprecate unused routes

RomanWlm:


luchenyuxx:
In general, I don't like the fact we need to deal with the prefix specifically in `deprecated`, even you are already in a prefixed context.

luchenyuxx:
```scala
class DeprecatedFilter extends SimpleFilter[Request, Response] with Logging {
  override def apply(request: Request, service: Service[Request,Response]): Future[Response] = {
    warn(s"Route ${request.path} is deprecated !!! Please refer to the documentation")
    service(request)
  }
}
    filter[DeprecatedFilter].get("/accounts") { request: AccountsRequest =>
      info(s"GET accounts $request")
      accountsService.accounts(request.walletInfo)
    }

```
Maybe something like this is better.

aymeric-ledger:
@RomanWlm Any advantage to use the `deprecated` from finatra instead of a plain log like @luchenyuxx proposed?
Otherwise we might as well just put a log inside the get request...

RomanWlm:
> ```scala
> class DeprecatedFilter extends SimpleFilter[Request, Response] with Logging {
>   override def apply(request: Request, service: Service[Request,Response]): Future[Response] = {
>     warn(s"Route ${request.path} is deprecated !!! Please refer to the documentation")
>     service(request)
>   }
> }
>     filter[DeprecatedFilter].get("/accounts") { request: AccountsRequest =>
>       info(s"GET accounts $request")
>       accountsService.accounts(request.walletInfo)
>     }
> ```
> 
> Maybe something like this is better.

👍 I completely not had look on Filter stuffs ! This is Definitely more elegant and better than WDController stuff then i'll propose something with filters!

@aymeric-ledger Unfortunately the deprecated is not from finatra :trollface: this my own undesired child, I would factorise the deprecated warn log code. I'll propose something with filter, it seems to be perfect for this :)

aymeric-ledger:
@RomanWlm Ok 😄 
Yea the filter solution is way more elegant.

PR #173, opened 2019-11-22
Bump lib-core to latest.

phaazon:


phaazon:
`HODL` that until I update it with:

- The commit hash from `lib-core` that contains the EC fixes.
- Jean-Michel.

PR #183, opened 2019-12-06
[Back-379] Batch erc20 balances to core lib

RomanWlm:
* Retrieve all token addresses and batch on erc20
  * Replace old getERC20Balance when it is possible
  * Single token get balance is using also batched method

/!\ **This PR is relying on this one https://github.com/LedgerHQ/djinni/pull/41 which fix a JNI issue. Please consider status it before merging this one.**


luchenyuxx:
So lib core has perf improve on batch get balance?

RomanWlm:
> So lib core has perf improve on batch get balance?

Absolutely batching getBalances trigger only **one** http call from core to explorer instead of one call **per balance**

PR #184, opened 2019-12-13
[Back-386] error and access logging

RomanWlm:
Enhance Logging when error occur 
Enable AccessLogger logs

### DIFF
--- a/src/main/scala/co/ledger/wallet/daemon/controllers/responses/ResponseSerializer.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/controllers/responses/ResponseSerializer.scala
@@ -10,33 +10,35 @@ object ResponseSerializer extends Logging {
     val msg = Map(
       "response"->s"Internal server error occurred while processing the request. ${caught.getMessage}"
     )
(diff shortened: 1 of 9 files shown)


src/main/scala/co/ledger/wallet/daemon/controllers/responses/ResponseSerializer.scala
```diff
@@ -1,31 +1,39 @@
...
-      .body(ErrorResponseBody(ErrorCode.Internal_Error, Map(
-        "response"->s"Internal server error occurred while processing the request. ${caught.getMessage}"
-      )))
+      .body(ErrorResponseBody(ErrorCode.Internal_Error, msg))
   }
 
-  def serializeBadRequest(msg: Map[String, Any], response: ResponseBuilder): Response = {
-    info(ErrorCode.Bad_Request)
+  def serializeBadRequest(request: Request, msg: Map[String, Any], response: ResponseBuilder): Response = {
+    log((str) => info(str), ErrorCode.Bad_Request, request, msg)
```
aymeric-ledger: Just pass `info` instead of `s => info(s)`

PR #185, opened 2019-12-19
Back 354 batch eth btc operations

RomanWlm:
Expose core pagination for btc and eth

### DIFF
--- a/src/it/scala/co/ledger/wallet/daemon/api/AccountsApiTest.scala
+++ b/src/it/scala/co/ledger/wallet/daemon/api/AccountsApiTest.scala
@@ -3,11 +3,16 @@ package co.ledger.wallet.daemon.api
 import java.util.UUID
 
 import co.ledger.core.TimePeriod
+import co.ledger.wallet.daemon.configurations.DaemonConfiguration
 import co.ledger.wallet.daemon.models.{AccountDerivationView, AccountView, FreshAddressView}
 import co.ledger.wallet.daemon.services.OperationQueryParams
 import co.ledger.wallet.daemon.utils.APIFeatureTest
-// import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.JsonNode
 import com.twitter.finagle.http.{Response, Status}
+import scala.collection.JavaConverters._
+
+import collection.mutable._
+import scala.collection.JavaConverters.collectionAsScalaIterableConverter
 
 class AccountsApiTest extends APIFeatureTest {
 
@@ -130,8 +135,6 @@ class AccountsApiTest extends APIFeatureTest {
     deletePool("account_pool")
   }
 
-  // FIXME: Broken test
-  /*
   test("AccountsApi#Get account operations") {
 
     def getUUID(field: String, content: Map[String, JsonNode]): Option[UUID] = {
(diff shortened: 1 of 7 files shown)


luchenyuxx:
I saw you are dropping the previous way of returning operations.
Just to let you know, the previous way of pagination is intended to keep the consistency when user query, e.g. new operation added during the pagination query.
If we use the new method provided by lib core, we need to make sure it fulfill the requirements or the user just don't care about the consistency.

RomanWlm:
> I saw you are dropping the previous way of returning operations.
> Just to let you know, the previous way of pagination is intended to keep the consistency when user query, e.g. new operation added during the pagination query.
> If we use the new method provided by lib core, we need to make sure it fulfill the requirements or the user just don't care about the consistency.

Operations provided by core lib are ordered by date, this should ensure consistency. Or may be iam missing something ? If we do not want to see new operations, we could add a consistency ts in order filter the new created operation. Thats a good question.

luchenyuxx:
The problem is that the offset won't be the intended offset if there is a new op come.
I don't think it could cause big problem, we just need to warn the user about it. The previous way avoid this by this complicated token cache.
By the way, I think these are quite equivalent:
```scala
    a.queryOperations().offset(offset).limit(batch).addOrder(OperationOrderKey.DATE, true).complete().execute().map(_.asScala.toList).map { coreOps =>
      val hashToERC20 = a.getOperations.asScala.toList.map(erc20Op => erc20Op.getHash -> erc20Op).toMap
      coreOps.map(coreOp => (coreOp, hashToERC20(coreOp.asEthereumLikeOperation().getTransaction.getHash)))
    }
```
```scala
account.operations(offset, batch, fullOp)
```

luchenyuxx:
@RomanWlm need test on user agent fix.

luchenyuxx:
> The problem is that the offset won't be the intended offset if there is a new op come.
> I don't think it could cause big problem, we just need to warn the user about it. The previous way avoid this by this complicated token cache.
> By the way, I think these are quite equivalent:
> Before:
> 
> ```scala
>     a.queryOperations().offset(offset).limit(batch).addOrder(OperationOrderKey.DATE, true).complete().execute().map(_.asScala.toList).map { coreOps =>
>       val hashToERC20 = a.getOperations.asScala.toList.map(erc20Op => erc20Op.getHash -> erc20Op).toMap
>       coreOps.map(coreOp => (coreOp, hashToERC20(coreOp.asEthereumLikeOperation().getTransaction.getHash)))
>     }
> ```
> 
> After:
> 
> ```scala
> account.operations(offset, batch, fullOp)
> ```
> 
> Could you tell the difference?

@RomanWlm the main problem here.

src/it/scala/co/ledger/wallet/daemon/api/AccountsApiTest.scala
```diff
@@ -167,6 +169,51 @@ class AccountsApiTest extends APIFeatureTest {
...
+    assertSyncPool(Status.Ok)
+    assertGetAccountOp(poolName, walletName, 0, "noexistop", 0, Status.NotFound)
+
+    // Check configuration limit is taken when no limit are provided
+    val opsAsJson_default = parse[Map[String, JsonNode]](assertGetBatchedAccountOps(poolName, walletName, 0, None, None, None, Status.Ok))
+    val ops_default = opsAsJson_default("operations")
+    assert(ops_default.size() == DaemonConfiguration.paginationOperationsLimit)
+
+    val opsAsJson_all = parse[Map[String, JsonNode]](assertGetBatchedAccountOps(poolName, walletName, 0, Some(0), Some(10), None, Status.Ok))
+    val ops_all = opsAsJson_all("operations").findValuesAsText("uid").asScala
```
ClaireGuerreGiordano: Is there a way to avoid Java to Scala collection conversions by using methods returning scala collections?

src/main/scala/co/ledger/wallet/daemon/clients/ScalaHttpClient.scala
```diff
@@ -72,7 +75,7 @@ class ScalaHttpClient(implicit val ec: ExecutionContext) extends co.ledger.core.
             }
           } while (size > 0)
           val data = outputStream.getBytes
-          if (isError) info(s"Received error ${new String(data)}")
+          if (isError) info(s"Received ${new String(data)}")
```
ClaireGuerreGiordano: Is it possible to use pattern matching over statusCode variable instead of if?

src/it/scala/co/ledger/wallet/daemon/api/AccountsApiTest.scala
RomanWlm: Unfortunately I think there is no straightforward way to get it directly as Scala value as JsonNode is coming from jackson which is a Java framework so it is not a big issue here, sifficient for our need at the moment.

PR #187, opened 2019-12-20
Back 404

luchenyuxx:
I found that when we have many accounts, the full synchronization will create dead lock in lib core. This is due to the parallelism.

The fix provided has three important change:

- make the synchronization of all accounts sequentially
- The timeout is moved to account synchronization level
- You won’t able to launch two full syncs at the same time

### DIFF
--- a/src/main/scala/co/ledger/wallet/daemon/modules/DaemonCacheModule.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/modules/DaemonCacheModule.scala
@@ -5,6 +5,7 @@ import java.util.concurrent.TimeUnit
 import co.ledger.wallet.daemon.async.MDCPropagatingExecutionContext.Implicits.global
 import co.ledger.wallet.daemon.configurations.DaemonConfiguration
 import co.ledger.wallet.daemon.database.{DaemonCache, DefaultDaemonCache}
+import co.ledger.wallet.daemon.exceptions.AccountSyncException
 import co.ledger.wallet.daemon.services.{PoolsService, UsersService}
 import com.google.inject.Provides
 import com.twitter.concurrent.NamedPoolThreadFactory
@@ -14,7 +15,7 @@ import javax.inject.Singleton
 
 import scala.concurrent.duration._
 import scala.concurrent.{Await, Future}
-import scala.util.{Failure, Success, Try}
+import scala.util.{Failure, Success}
 
 object DaemonCacheModule extends TwitterModule {
 
@@ -32,21 +33,22 @@ object DaemonCacheModule extends TwitterModule {
     val poolsService = injector.instance[PoolsService](classOf[PoolsService])
 
     def synchronizationTask(): Unit = {
-      val t0 = System.currentTimeMillis()
-      Try(Await.result(poolsService.syncOperations, 5.minutes)) match {
-        case Success(result) =>
-          result.foreach { r =>
+      try {
+        Await.result(poolsService.syncOperations, 30.minutes).foreach{
+          case Success(r) =>
             if (r.syncResult) {
               info(s"Synchronization complete for $r")
             }
             else {
               warn(s"Failed synchronizing $r")
             }
-          }
-          val t1 = System.currentTimeMillis()
-          info(s"Synchronization finished, elapsed time: ${t1 - t0} milliseconds")
-        case Failure(e) =>
-          error("Synchronization failed with exception", e)
+          case Failure(e: AccountSyncException) =>
+            error(e.getMessage, e)
+          case Failure(t) =>
+            error("Failed to synchronize account due to unknown exception", t)
+        }
+      } catch {
+        case t: Throwable => error("The full synchronization timed out in 30 minutes", t)
       }
     }
 
@@ -56,10 +58,10 @@ object DaemonCacheModule extends TwitterModule {
         threadFactory = new NamedPoolThreadFactory("scheduler-thread-pool")
       )
       scheduler.schedule(
-        Time.fromSeconds(DaemonConfiguration.synchronizationInterval._1),
-        Duration(DaemonConfiguration.synchronizationInterval._2, TimeUnit.HOURS))(synchronizationTask())
-      info(s"Scheduled synchronization job: initial start in ${DaemonConfiguration.synchronizationInterval._1} seconds, " +
-        s"interval ${DaemonConfiguration.synchronizationInterval._2} hours")
+        Time.fromSeconds(DaemonConfiguration.Synchronization.initialDelay),
+        Duration(DaemonConfiguration.Synchronization.interval, TimeUnit.HOURS))(synchronizationTask())
+      info(s"Scheduled synchronization job: initial start in ${DaemonConfiguration.Synchronization.initialDelay} seconds, " +
+        s"interval ${DaemonConfiguration.Synchronization.interval} hours")
     }
 
     val usersService = injector.instance[UsersService](classOf[UsersService])
--- a/src/main/scala/co/ledger/wallet/daemon/modules/DaemonCacheModule.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/modules/DaemonCacheModule.scala
@@ -34,7 +34,7 @@ object DaemonCacheModule extends TwitterModule {
 
     def synchronizationTask(): Unit = {
       try {
-        Await.result(poolsService.syncOperations, 30.minutes).foreach{
+        Await.result(poolsService.syncOperations, 1.hour).foreach{
           case Success(r) =>
             if (r.syncResult) {
               info(s"Synchronization complete for $r")
--- a/src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
@@ -5,7 +5,7 @@ import java.util.concurrent.atomic.AtomicBoolean
 import co.ledger.wallet.daemon.async.MDCPropagatingExecutionContext.Implicits.global
 import co.ledger.wallet.daemon.database.DaemonCache
 import co.ledger.wallet.daemon.database.DefaultDaemonCache.User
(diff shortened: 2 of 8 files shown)


src/main/scala/co/ledger/wallet/daemon/modules/DaemonCacheModule.scala
```diff
@@ -32,21 +32,16 @@ object DaemonCacheModule extends TwitterModule {
...
-      Try(Await.result(poolsService.syncOperations, 5.minutes)) match {
-        case Success(result) =>
-          result.foreach { r =>
-            if (r.syncResult) {
-              info(s"Synchronization complete for $r")
-            }
-            else {
-              warn(s"Failed synchronizing $r")
-            }
+      Await.result(poolsService.syncOperations, 30.minutes).foreach{
```
aymeric-ledger: This will throw if the timeout is reached (and other edge cases). Is it really what you want?

luchenyuxx: 👍

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
```diff
@@ -31,16 +36,56 @@ class PoolsService @Inject()(daemonCache: DaemonCache) extends DaemonService {
...
+    * We synchronize account by account in a synchronous way, to
+    * avoid dead lock of lib core. The test reveal that if we make
+    * parallel synchronization, the lib core will be locked.
+    *
     * @return a Future of sequence of result of synchronization.
     */
-  def syncOperations: Future[Seq[SynchronizationResult]] =
-    daemonCache.getUsers.flatMap { us =>
-      Future.sequence(us.map { user => user.sync()}).map (_.flatten)
+  def syncOperations: Future[Seq[Try[SynchronizationResult]]] = {
```
aymeric-ledger: Any reason to you Future[Seq[Try[_]]] instead of Future[Seq[_]] here?

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
```diff
@@ -31,16 +36,56 @@ class PoolsService @Inject()(daemonCache: DaemonCache) extends DaemonService {
...
+    import co.ledger.wallet.daemon.models.Account._
+    import co.ledger.wallet.daemon.models.Wallet._
+
+    import scala.concurrent.duration._
```
aymeric-ledger: This should go in top imports.

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
luchenyuxx: yes, `Try` catch the timeout of account sync

src/main/scala/co/ledger/wallet/daemon/modules/DaemonCacheModule.scala
luchenyuxx: Emm, I think I would throw the Exception than catch it.

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
aymeric-ledger: I mean are you actually using it, because otherwise you should flatten this all in a `Future[Seq[_]]`.

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
```diff
@@ -31,16 +36,56 @@ class PoolsService @Inject()(daemonCache: DaemonCache) extends DaemonService {
...
+            }
+        }
+      }
+      resultFuture.onComplete(_ => syncOnGoing.set(false))
+      resultFuture
     }
+  }
+
+  // To avoid launching sync at the same time
+  private val syncOnGoing = new AtomicBoolean(false)
```
aymeric-ledger: We do have multiple instances for WD right?

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
luchenyuxx: yeah, it's used on all the usage of this function

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
luchenyuxx: no, we have a endpoint to resync all. just to avoid someone access the endpoint endlessly.

src/main/scala/co/ledger/wallet/daemon/modules/DaemonCacheModule.scala
```diff
@@ -32,21 +32,16 @@ object DaemonCacheModule extends TwitterModule {
...
+            info(s"Synchronization complete for $r")
+          }
+          else {
+            warn(s"Failed synchronizing $r")
           }
-          val t1 = System.currentTimeMillis()
-          info(s"Synchronization finished, elapsed time: ${t1 - t0} milliseconds")
         case Failure(e) =>
-          error("Synchronization failed with exception", e)
+          error(e.getMessage, e)
```
RomanWlm: We should keep a fixed message here in order to quickly locate into logs this case

src/main/scala/co/ledger/wallet/daemon/modules/DaemonCacheModule.scala
aymeric-ledger: That's so strange of you to say that 😂 But ok 👍

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
RomanWlm: Good, we should refuse sync from http calls if one is ongoing. Good catchup.

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
aymeric-ledger: Ok I see, you match on the Try result for logging. Altho you could do it right here instead of returning a `Seq[Try[_]]`, it would make more sense to me.

src/main/scala/co/ledger/wallet/daemon/modules/DaemonCacheModule.scala
RomanWlm: Synchronization failed with exception : e.getMessage

aymeric-ledger: Keep a fix header !

luchenyuxx: the message is provided in the AccountSyncException

luchenyuxx: emmm, wait, you are right, `e` is not necessarily a `AccountSyncException`

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
aymeric-ledger: Yes, sad state of things but good catch 👍

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
luchenyuxx: Not just for logging, the resync all endpoint also need a list of synchronization results.

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
```diff
@@ -31,16 +39,52 @@ class PoolsService @Inject()(daemonCache: DaemonCache) extends DaemonService {
...
+  def syncOperations: Future[Seq[Try[SynchronizationResult]]] = {
+    if (syncOnGoing.get()) {
+      Future.failed(new Exception("Synchronization is on going"))
+    } else {
+      syncOnGoing.set(true)
+      val accountsFuture = for {
+        users <- daemonCache.getUsers
+        pools <- Future.sequence(users.map(_.pools())).map(_.flatten)
+        wallets <- Future.sequence(pools.map { p =>
+          for {
```
aymeric-ledger: Would be more readable with a simple p.map(_.wallets.map(p.name, _))

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
```diff
@@ -31,16 +39,52 @@ class PoolsService @Inject()(daemonCache: DaemonCache) extends DaemonService {
...
+            wallets <- p.wallets
+          } yield wallets.map((p.name, _))
+        }).map(_.flatten)
+        accounts <- Future.sequence(wallets.map { case (poolName, w) =>
+          for {
```
aymeric-ledger: Same comment

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
luchenyuxx: actually it won't :trollface: 
I came from there
it actually is `p.map(p => p.wallets.map(ws => ws.map(w => (p.name, w))))

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
luchenyuxx: same comment

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
aymeric-ledger: With some indentation I prefer the .map version 🙈

aymeric-ledger: I just find for comprehension when there is no flatMap more confusing.

luchenyuxx:
I made a force push to rebase on develop

PR #193, opened 2019-12-30
Full sync operations: timer and non blocking

nukky94:
The goal of this PR is to use built-in timer logger from finatra and provide a non blocking timeoutable and cancellable futures for sync operations

### DIFF
--- a/src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
@@ -1,6 +1,7 @@
 package co.ledger.wallet.daemon.services
 
-import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.{Executors, ThreadFactory}
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicLong}
 
 import co.ledger.wallet.daemon.async.MDCPropagatingExecutionContext.Implicits.global
 import co.ledger.wallet.daemon.database.DaemonCache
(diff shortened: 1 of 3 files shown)


src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
```diff
@@ -69,14 +71,33 @@ class PoolsService @Inject()(daemonCache: DaemonCache) extends DaemonService {
...
       } yield accounts
-      val resultFuture = accountsFuture.map { accounts =>
-        accounts.map {
-          case (poolName, walletName, a) =>
-            val f = a.sync(poolName, walletName)
-            Try(Await.result(f, 30.minute)).recoverWith{ case t =>
-              Failure(AccountSyncException(poolName, walletName, a.getIndex, t))
+
+      // custom ec to avoid thread starvation in the global ec
+      val fullSyncEc: ExecutionContext = ExecutionContext.fromExecutor(
```
luchenyuxx: not sure about the purpose of this ec. the sync won't happen in global ec, they are directed to lib core ec.

src/main/scala/co/ledger/wallet/daemon/utils/FutureUtils.scala
```diff
@@ -0,0 +1,79 @@
...
+  /**
+    * Scala Future doesn't provide a cancel function.
+    * As a workaround, you could use the firstCompletedOf to wrap 2 futures.
+    * The future you want to cancel and a future that comes from a custom Promise.
+    * You could then cancel the thus created future by failing the promise.
+    * @param future Caller passes a future to execute
+    * @param onError Error handler
+    * @return Future[T]
+    */
+  def cancellable[T](future: Future[T])(onError: => Unit)(implicit ec: ExecutionContext): (() => Unit, Future[T]) = {
```
luchenyuxx: Just not that the `future` is not cancelled after the `cancellable` is called. It's an improper name

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
nukky94: The goal here is to use a different ec for timeoutable operation:
https://imgur.com/a/AbMJZpH
But if we look at it closer, it still uses the global ec.
Should I correct that or revert it?

src/main/scala/co/ledger/wallet/daemon/utils/FutureUtils.scala
nukky94: I see, what do you propose? `failFast` ?

src/main/scala/co/ledger/wallet/daemon/services/PoolsService.scala
luchenyuxx: I mean the actual sync won't happen in global ec even if you remove this ec.
They are happening in the dedicated lib core ec.
So what your wrote is not necessary. You can just remove it.

src/main/scala/co/ledger/wallet/daemon/utils/FutureUtils.scala
luchenyuxx: No, it's fine.

PR #195, opened 2020-01-09
Fix high memory consumption on the accounts controller.

phaazon:
# Content

That PR has several implications:

- The associated operations counts, not needed (according to @malikfassi), are removed.
  The interface changed to just get balances.
- The memory consumption is now flatten.

Relates and fixes LLC-485, BACK-415.

# Mandatory picture of animal


RomanWlm:
> We should just be careful that Gate is not using this anymore !

Absolutely, we will need a sync with vault regarding the gate 👍

PR #196, opened 2020-01-09
BACK-407 : ERC20NotFound are raised when there is no token anymore on the account

RomanWlm:
Little PR for smoother behavior when there is no token on an account.

Return empty operation list or null balance when there is no ERC20 found : token balance, balance history and operations

We log for audit purpose when the situation occur.

This should enhance a bit the accuracy of failure report from monitoring tools.


luchenyuxx:
better to have some tests

RomanWlm:
> better to have some tests

I infinitively agree. I would deliver it with tests and check that we are not returning 5xx HTTP errors thanks to this change but I discovered the ehtereum like account are not tested. I had to create account and complete test dataset to achieve this. This PR is an hotfix, I would do better but cannot deliver it on time. The test coverage issue is tracked on our Jira and will be the very nexts handled topics

PR #197, opened 2020-01-13
BACK-421 : Balance history - Bug fix Return a 0 filled balance

RomanWlm:
BACK-421 : Balance history - Bug fix
Return a 0 filled balance list even if there is no tokens or operations on the account instead of empty list.


RomanWlm:
> Could we also add integration test?

@luchenyuxx Absolutely, this case (and ERC in general) are not covered and it MUST be covered ASAP. If our coverage was sufficient, actually, this bug would not exists at all :) But It's still an ERC20 service and I have to deal with legacy => https://github.com/LedgerHQ/ledger-wallet-daemon/pull/196#issuecomment-572973721

PR #198, opened 2020-01-20
BACK-431 : Retry policy, short living reusable connections and logs

RomanWlm:


src/main/scala/co/ledger/wallet/daemon/configurations/DaemonConfiguration.scala
```diff
@@ -15,6 +15,12 @@ object DaemonConfiguration {
   private val DEFAULT_AUTH_TOKEN_DURATION: Int = 3600 * 1000 // 30 seconds
   private val DEFAULT_SYNC_INTERVAL: Int = 24 // 24 hours
   private val DEFAULT_SYNC_INITIAL_DELAY: Int = 300 // 5 minutes
+  // Default value for keeping alive connections inside the fallback pool
+  private val DEFAULT_FALLBACK_CONNECTION_TTL_SECONDS: Int = 120 // 120 seconds
```
luchenyuxx: In fact, these configs are used by all http clients, not only the fallback one.

RomanWlm: Ok, I'll be make this more clear on config and naming

PR #199, opened 2020-01-21
Integrating libcore with PostgreSQL support

KhalilBellakrid:
This related to: https://ledgerhq.atlassian.net/browse/BACK-417

**Note:** we will remove the HODL label after performing few benchmarks.


RomanWlm:
We could update the documentation regarding the postgresql setup in order to up locally a wallet daemon instance with pg support

src/main/resources/application.conf.sample
```diff
@@ -7,6 +9,13 @@ sqlite3 = {
   connectionPool = disabled
 }
 
+postgres = {
```
RomanWlm: As discussed : Explicit naming for core related configurations

src/main/resources/application.conf.sample
```diff
@@ -7,6 +9,13 @@ sqlite3 = {
...
+    url = "postgres://localhost:5432"
+    url = ${?PG_URL}
```
dud225: it would be great to prefix this variable and the one below by `CORE_` to clarify its field of application.

RomanWlm: As discussed we would add more configuration stuffs : 

`
postgres = {
  host = "localhost"
  host = ${?CORE_PG_HOST}
  port = "5432"
  port = ${?CORE_PG_PORT}
  user = "postres"
  user = ${?CORE_PG_USER}
  pwd = ${?CORE_PG_PWD}
  pool_size = 30
  pool_size = ${?PG_CONNECTION_POOL_SIZE_PER_WALLET_POOL}
}`

PR #203, opened 2020-01-30
Add configuration to use custom explorer url and version

KhalilBellakrid:
This is related to https://ledgerhq.atlassian.net/browse/BACK-458.
We provide a way to choose a custom configuration for BTC explorers.


KhalilBellakrid:
> It seems that the only config need to add is the `explorer_version`.
> The `explorer_api` is the same as `host`

Actually it is done in purpose because `explorer.api.paths.host` is used in other parts (e.g. fallback), so not to introduce a regression on those parts we preferred to keep the previous one and add a custom url to be used only for the purpose of customization.
@luchenyuxx @RomanWlm WDYT? Or are we sure that `explorer.api.paths.host` using is not impacting other behaviours ?

luchenyuxx:
@KhalilBellakrid The `host` is also used to get fees.
IMO, the good way to avoid regression is to have tests, not compromising code quality.

RomanWlm:
> @KhalilBellakrid The `host` is also used to get fees.
> IMO, the good way to avoid regression is to have tests, not compromising code quality.

Totally agree with this. Let's consider this as a potential hotfix that we could deploy only if we have to and implement properly under test coverage the clean solution. Let me know whet you think, I'll do the rest.

KhalilBellakrid:
> @KhalilBellakrid The `host` is also used to get fees.
> IMO, the good way to avoid regression is to have tests, not compromising code quality.

@luchenyuxx I totally agree, but here we are in the context of a hotfix, that's why we went with avoiding any impact.

@RomanWlm could you please open a ticket regarding that? Thanks in advance for that ! 🙏

luchenyuxx:
@KhalilBellakrid The impact is already identified. The path to a better code base is clear. Please take time to make it work in a good way.
Creating a ticket and changing it back afterwards is just creating more work which impact also infra, since it's config related.

src/main/resources/application.conf.sample
```diff
@@ -84,6 +84,14 @@ explorer = {
         host = "api.ledgerwallet.com"
         port = 80
       }
+      {
+        currency = bitcoin
+        host = "api.ledgerwallet.com"
+        host = ${?WALLET_BTC_EXPLORER_ENDPOINT}
+        port = 80
+        port = ${?WALLET_BTC_EXPLORER_PORT}
+        custom_explorer_version = ${?WALLET_BTC_CUSTOM_EXPLORER_VERSION}
```
RomanWlm: Non blocking remark : I would name this variable just `explorer_version` and consider v2 as default value when this optional option is not provided

KhalilBellakrid: @RomanWlm I will do that !

PR #204, opened 2020-01-30
BACK-454 : Add Account data cleaning endpoint

RomanWlm:
Currently the following scenario is not working : 
Create account ✅ 
Check there is no operation ✅ 
Sync  ✅
Check that we retrieve operations  ✅
Delete op  ✅
Check there is no operations any more ✅
Sync ✅
Check operations are back ⛔️  => Still no operations

I'am suspecting some state are still persisted on db, like last block synchronized, date of last sync or something like this. Then the new sync does not bring back again operations.
@phaazon @KhalilBellakrid  Can you confirm this case should works from libcore point of view please or if I missed something ?


phaazon:
It’s possible it’s linked to the state we persist whenever we do a sync, yes. We need to do a quick review of the abstract sync code, I’m pretty sure it’s easy to spot whether we clean the sync object when we delete data — I guess we don’t.

RomanWlm:
Currently the following scenario is not working : 
Create account ✅ 
Check there is no operation ✅ 
Sync  ✅
Check that we retrieve operations  ✅
Delete op  ✅
Check there is no operations any more ✅
Sync ✅
Check operations are back  ✅ 

Now tests are green since we've updated libcore version.

PR #207, opened 2020-02-12
Back 474 update default config, manage tls based on url protocol

RomanWlm:
Hotfix, this code will be moved and refactored when the work done on ScalaHttpClient will be available


src/it/resources/application.conf
```diff
@@ -56,47 +56,50 @@ whitelist = [
...
         currency = default
-        host = "api.ledgerwallet.com"
-        port = 80
+        host = "https://api.ledgerwallet.com"
+        port = 443
       }
       {
         currency = bitcoin
-        host = "https://bitcoin-mainnet.explorers.dev.aws.ledger.fr"
+        host = "https://api.ledgerwallet.com"
```
KhalilBellakrid: We chose v3 explorers intentionally to test a potential switch towards BTC v3 as a fallback, I think we should keep that usage in tests.

PR #210, opened 2020-02-21
Back-475 : Make run our it tests behind squid proxy and with a PG instance with ssl enabled 

RomanWlm:


src/it/resources/application.conf
```diff
@@ -1,18 +1,38 @@
+database_engine = ${?DATABASE_ENGINE}
```
luchenyuxx: I think this should be put after the default value to be able to override it.

PR #211, opened 2020-02-24
Back-501 : list utxos for BTC Accounts

RomanWlm:
Expose the ability to list utxo from BTC accounts
- Test are not complete as there is no available account ready for this scenario on tests (ie with utxo on it) need to integrate an account with utxos
- The only test done for now is check that service is returning 200 http code
- Number of confirmation are not available yet so we are returning 0 for now

Usage Exemple :
`/pools/:pool_name:/wallets/:wallet_name:/accounts/:idx:/utxo?offset=3&batch=5`

Where _offset_ is the index of utxo from which we want to list and _batch_ the maximum size of returned list. If non of them are provided default values are _offset=0_ & _batch=Integer.MAX_VALUE_

### DIFF
--- a/src/main/scala/co/ledger/wallet/daemon/controllers/AccountsController.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/controllers/AccountsController.scala
@@ -150,6 +150,11 @@ class AccountsController @Inject()(accountsService: AccountsService) extends Con
         accountsService.synchronizeAccount(request.accountInfo)
       }
 
+      // List of utxos available on this account
+      get("/utxo") { request: UtxoAccountRequest =>
+        accountsService.getUtxo(request.accountInfo, request.offset, request.batch)
+      }
+
       // List of tokens on this account
       get("/tokens") { request: AccountRequest =>
         accountsService.getTokenAccounts(request.accountInfo)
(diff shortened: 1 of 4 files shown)


src/main/scala/co/ledger/wallet/daemon/controllers/AccountsController.scala
```diff
@@ -150,9 +150,15 @@ class AccountsController @Inject()(accountsService: AccountsService) extends Con
         accountsService.synchronizeAccount(request.accountInfo)
       }
 
+      // List of tokens on this account
```
luchenyuxx: unrelated comment

src/main/scala/co/ledger/wallet/daemon/controllers/AccountsController.scala
```diff
@@ -150,9 +150,15 @@ class AccountsController @Inject()(accountsService: AccountsService) extends Con
...
+      get("/utxo") { request: UtxoAccountRequest =>
+        accountsService.getUtxo(request.accountInfo, request.offset, request.batch)
+      }
+
       // List of tokens on this account
       get("/tokens") { request: AccountRequest =>
         accountsService.getTokenAccounts(request.accountInfo)
+
```
luchenyuxx: one more empty line

src/main/scala/co/ledger/wallet/daemon/models/Account.scala
```diff
@@ -1,5 +1,6 @@
 package co.ledger.wallet.daemon.models
 
+import java.util.ArrayList
```
luchenyuxx: Avoid using java list. you can use `asScala` from `import scala.collection.JavaConverters._`

luchenyuxx: To convert ArrayList to scala List

PR #216, opened 2020-03-12
Add UTXO count to response when querying UTXOs on Account controller

KhalilBellakrid:
This is related to : https://ledgerhq.atlassian.net/browse/BACK-528


src/main/scala/co/ledger/wallet/daemon/controllers/AccountsController.scala
```diff
@@ -152,7 +152,9 @@ class AccountsController @Inject()(accountsService: AccountsService) extends Con
 
       // List of utxos available on this account
       get("/utxo") { request: UtxoAccountRequest =>
-        accountsService.getUtxo(request.accountInfo, request.offset, request.batch)
+        for {
+          utxoResponse <- accountsService.getUtxo(request.accountInfo, request.offset, request.batch)
+        } yield UtxoAccountResponse(utxoResponse._1, utxoResponse._2)
```
luchenyuxx:
```suggestion
        accountsService.getUtxo(request.accountInfo, request.offset, request.batch)
           .map(utxoResponse => UtxoAccountResponse(utxoResponse._1, utxoResponse._2))
```

KhalilBellakrid:
> do we have the corresponding lib core in the commit?

This is already available on used libcore so no need for any other action.

PR #225, opened 2020-03-20
[COIN-240] Bech32 support

nukky94:


src/it/resources/application.conf
```diff
@@ -101,7 +101,7 @@ explorer = {
         host = ${?WALLET_BTC_TESTNET_EXPLORER_ENDPOINT}
         port = 443
         port = ${?WALLET_BTC_TESTNET_EXPLORER_PORT}
-        explorer_version = "v3"
+        explorer_version = "v2"
```
RomanWlm: We have to move to V2 as V3 is not managing segwit ?

src/main/scala/co/ledger/wallet/daemon/configurations/DaemonConfiguration.scala
```diff
@@ -25,6 +25,8 @@ object DaemonConfiguration extends Logging{
   private val DEFAULT_CLIENT_CONNECTION_RETRY_BACKOFF: Int = 50 // Linear Backoff policy delta ms
   private val DEFAULT_CORE_POOL_THREADS_FACTOR: Int = 4 // Default value for allocated threads per pool for core lib
 
+  val SUPPORTED_NATIVE_SEGWIT_CURRENCIES = Seq("bitcoin", "bitcoin_testnet", "litecoin", "qtum")
```
RomanWlm: Upper case for default values only.
Every values here should be configurable or somewhere else.

In my opinion this should be managed by the libcore.
Is_native_segwit part of a wallet config object that is bring up to the libcore.

src/main/scala/co/ledger/wallet/daemon/exceptions/Exceptions.scala
```diff
@@ -19,6 +19,12 @@ case class InvalidEIP55Format(address: String)
     val code = ErrorCodes.CORE_BAD_REQUEST
   } with DaemonException(msg)
 
+case class InvalidCurrencyForNativeSegwit(currencyName: String)
+  extends {
+    val msg = s"Native segwit not supported for currency $currencyName"
```
RomanWlm:
```suggestion
    val msg = s"Native SegWit is not supported for currency $currencyName"
```

src/main/scala/co/ledger/wallet/daemon/models/Pool.scala
```diff
@@ -250,25 +257,38 @@ class Pool(private val coreP: core.WalletPool, val id: Long) extends Logging {
...
-    val apiUrl = DaemonConfiguration.explorer.api.paths.get(currencyName) match {
-      case Some(path) =>
-        path.explorerVersion match {
-          case Some(version) =>
-            walletConfig.putString("BLOCKCHAIN_EXPLORER_VERSION", version)
-          case _ =>
-        }
-        new URL(s"${path.host}:${path.port}").toString
-      case None => ConfigurationDefaults.BLOCKCHAIN_DEFAULT_API_ENDPOINT
+  private def buildWalletConfig(currency: core.Currency, isNativeSegwit: Boolean): Future[core.DynamicObject] = {
```
luchenyuxx: prefer Either than Future if no future happens

RomanWlm:
Good job !! 
Some renaming and minor changes to do.
Can you also clean the git commit history : squash intermediate commits (like `Rename bech32 to native segwit`) and also put the ticket that the PR solve in front of commits names ? `COIN-240 : `

src/it/resources/application.conf
nukky94: https://api.ledgerwallet.com/blockchain/v3/btc_testnet/blocks/current
v3 doesn't seem to work, I use the v2 version as defined for bitcoin.
I also apply this change in `application.conf.sample`

src/main/scala/co/ledger/wallet/daemon/configurations/DaemonConfiguration.scala
nukky94: 👍 for configurable list.
The purpose of this list is to support only wanted currencies, otherwise, all coins which are in bitcoin family have native segwit support

src/it/resources/application.conf
RomanWlm: yes actually I checked and explorer V3 for btc testnet is not available yet, so It has been moved to V2 by default for the moment 👍

src/main/scala/co/ledger/wallet/daemon/exceptions/Exceptions.scala
RomanWlm: 👍

PR #230, opened 2020-03-26
Update the libcore version to 3.3.1

RomanWlm:


KhalilBellakrid:
This is done to patch only https://github.com/LedgerHQ/lib-ledger-core/pull/536 on top of already used libcore !

PR #239, opened 2020-04-16
[Back 552] Enhance test cases : ETH / ERC20 / XRP Account Sync Histo Balance

RomanWlm:
Bring test coverage from 68.90% to 76.43% 🐙


.circleci/config.yml
```diff
@@ -89,6 +89,7 @@ jobs:
             CORE_PG_PWD: ledger_secret
             CORE_PG_DB_NAME_PREFIX: wdtest_
             CORE_PG_CONNECTION_POOL_SIZE_PER_WALLET_POOL: 2
+            SBT_OPTS: -Xmx3200m -Xss2M -XX:MaxMetaspaceSize=1024M
```
luchenyuxx, line 92: Xss to 2m is too small,
the default is 128m if I recall correctly

src/main/scala/co/ledger/wallet/daemon/models/Pool.scala
```diff
@@ -278,10 +278,13 @@ class Pool(private val coreP: core.WalletPool, val id: Long) extends Logging {
               walletConfig.putString("BLOCKCHAIN_EXPLORER_VERSION", version)
             case _ =>
           }
-          new URL(s"${path.host}:${path.port}").toString
-        case None => ConfigurationDefaults.BLOCKCHAIN_DEFAULT_API_ENDPOINT
+          new URL(s"${path.host}:${path.port}")
+        case None => new URL(s"${ConfigurationDefaults.BLOCKCHAIN_DEFAULT_API_ENDPOINT}:433")
```
luchenyuxx, line 282: should not hard code.

RomanWlm: Thanks !! Good catch, this has been commited by mistake

.circleci/config.yml
RomanWlm: Xss is the stack size, I think default is 4M, I'll remove this setting

PR #243, opened 2020-04-30
[Back-582] fees harmonization across coins 

RomanWlm:
Harmonization of Create transaction payloads

- All BTC/ETH/XRP create transaction payload has `fees_level` field which is an enum value {`SLOW`, `NORMAL`, `FAST`}. 
- User can **override** fees by setting `fees` field on payload.
- Validation methods are ensuring that at least one of them is setup and will reject any request without at least fees or fees_level filled.
 
______________________

The fees level defines the amount of fees used according the current network load : 

**BTC** : 
- _SLOW_ : returned by explorer, amount to be included in the nexts 6 blocks 
- _NORMAL_ : returned by explorer, amount to be included in the nexts 3 blocks 
- _FAST_ : returned by explorer, amount to be included in the nexts 2 blocks 

**ETH** : Network `Gas Price * ETH_XXX_FEES_FACTOR` (configurable)
- _SLOW_ : Network Gas Price * 0.75 (Default conf value)
- _NORMAL_ : Network Gas Price * 1 (Default conf value)
- _FAST_ : Network Gas Price * 1.25 (Default conf value)

**XRP** : Note that `load coast`, `open ledger coast` and `base fee` are defined by the network 
- _SLOW_ : base fee
- _NORMAL_ : Base fee + load coast 
- _FAST_ :  base fee + (loadCoast + openLedgerCost) * 1.5

### DIFF
--- a/src/main/scala/co/ledger/wallet/daemon/clients/ApiClient.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/clients/ApiClient.scala
@@ -4,9 +4,10 @@ import java.net.URL
 
 import co.ledger.core
 import co.ledger.wallet.daemon.configurations.DaemonConfiguration
+import co.ledger.wallet.daemon.configurations.DaemonConfiguration._
 import co.ledger.wallet.daemon.models.FeeMethod
-import co.ledger.wallet.daemon.utils.{HexUtils, NetUtils}
 import co.ledger.wallet.daemon.utils.Utils._
+import co.ledger.wallet.daemon.utils.{HexUtils, NetUtils}
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.twitter.finagle.http.{Method, Request}
 import com.twitter.finatra.json.FinatraObjectMapper
@@ -45,7 +46,7 @@ class ApiClient(implicit val ec: ExecutionContext) extends Logging {
       mapper.objectMapper.readTree(response.contentString)
         .fields.asScala.filter(_.getKey forall Character.isDigit)
         .map(_.getValue.asInt).toList.sorted.map(BigInt.apply) match {
-        case low :: medium :: high :: Nil => FeeInfo(high, medium, low)
+        case low :: medium :: high :: Nil => BtcFeeInfo(high, medium, low)
         case _ =>
           warn(s"Failed to retrieve fees from explorer, falling back on default fees.")
           defaultBTCFeeInfo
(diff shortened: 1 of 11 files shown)


src/main/scala/co/ledger/wallet/daemon/clients/ApiClient.scala
```diff
@@ -37,15 +38,15 @@ class ApiClient(implicit val ec: ExecutionContext) extends Logging {
   def fallbackService(currency: String): Option[(URL, ScalaHttpClientPool)] =
     currencyServiceURLFor(currency).fallback.map(url => (url, fallbackServices))
 
-  def getFees(currencyName: String): Future[FeeInfo] = {
+  def getFees(currencyName: String): Future[BtcFeeInfo] = {
```
ClaireGuerreGiordano, line 41: This method could be named getFeesBtc

RomanWlm: Good catch. This method is intended to get fees for all coins. Currently we are managing only BTCLike fees by this way. I moved to FeeInfo trait for type returned.

src/main/scala/co/ledger/wallet/daemon/models/Account.scala
```diff
@@ -273,8 +274,8 @@ object Account extends Logging {
         case None => ethereumTransactionBuilder(ti, a, c)
       }
       gasPrice <- ti.gasPrice match {
-        case Some(amount) => Future.successful(amount)
-        case None => ClientFactory.apiClient.getGasPrice(c.getName)
+        case Some(gPrice) => Future.successful(gPrice)
```
malikfassi, line 277: Shouldnt you use `ClientFactory.apiClient.getFees(c.getName).map(f => f.getAmount(ti.feesSpeedLevel.getOrElse(FeeMethod.NORMAL)))` to get `gas_price` ?

src/main/scala/co/ledger/wallet/daemon/models/Account.scala
```diff
@@ -326,10 +327,14 @@ object Account extends Logging {
     ti.destinationTag.foreach(builder.setDestinationTag)
 
     for {
-      fees <- ti.fees.fold(ClientFactory.apiClient.getFeesRipple)(Future.successful)
+      feeMethod <- Future.successful(ti.feesSpeedLevel.getOrElse(FeeMethod.NORMAL))
```
malikfassi, line 330: here we should be able to create a ripple tx with hardcoded fees aswell.

malikfassi: nvm 🤦

src/main/scala/co/ledger/wallet/daemon/models/Account.scala
malikfassi: nvm 🤦

src/main/scala/co/ledger/wallet/daemon/models/Account.scala
RomanWlm: ✔️

src/main/scala/co/ledger/wallet/daemon/models/Account.scala
RomanWlm: ✔️

PR #250, opened 2020-06-18
Back-646 : Near Real Time notification for account states

RomanWlm:


src/main/scala/co/ledger/wallet/daemon/exceptions/Exceptions.scala
```diff
@@ -132,6 +137,7 @@ object ErrorCodes {
   val INVALID_CURRENCY_FOR_ERC20 = 210
   val ACCOUNT_SYNC_FAILED = 211
   val SYNC_ON_GOING = 212
+  val RESYNC_ON_GOING = 213
```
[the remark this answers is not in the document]
luchenyuxx: nice catch

src/main/scala/co/ledger/wallet/daemon/services/RabbitMQ.scala
```diff
@@ -0,0 +1,31 @@
...
+    info(s"RabbitMQ: connected to ${conn.getAddress.toString}")
+    conn
+  }
+  private val chan = conn.createChannel()
+  private val declaredExchanges = mutable.HashSet[String]()
+
+  // Channel is not recommended to be used concurrently, hence synchronized
+  def publish(exchangeName: String, routingKeys: List[String], payload: Array[Byte]): Unit = synchronized {
+    if (!declaredExchanges.contains(exchangeName)) {
+      chan.exchangeDeclare(exchangeName, BuiltinExchangeType.TOPIC)
```
dkremer-ledger, line 25: you should add 
`chan.exchangeDeclare(exchangeName, BuiltinExchangeType.TOPIC, true)`
because the exchange is durable actually.

src/main/resources/application.conf.sample
```diff
@@ -55,9 +55,20 @@ h2mem1 = {
 
 realtimeobservation = true
 
+# Configuration for Account Synchronizer Manager (ASM)
 synchronization = {
-  initial_delay_in_seconds = 120
-  interval_in_hours = 1
+  initial_delay_in_seconds = 120   # Not used
+  interval_in_hours = 1            # Not used
```
RomanWlm, line 61: If not used, we should remove them, isn't it ?

PR #258, opened 2020-08-04
Back 800 update configuration

RomanWlm:
Refresh configuration of WD
- Default Vault Explorer V3 
- update_wallet_config is not an option anymore (reload at startup)
- disable sync token by default


mbasire-ledger (changes requested):

src/it/resources/application.conf
```diff
@@ -134,9 +134,9 @@ explorer = {
       }
       {
         currency = ethereum_classic
-        host = "https://explorers.api.live.ledger.com"
+        host = "https//etc.explorers.vault.ovh.ppr.ldg-tech.com"
```
mbasire-ledger: `host = "https://etc.explorers.vault.ovh.ppr.ldg-tech.com"`

mbasire-ledger (approved):

src/it/resources/application.conf
```diff
@@ -139,7 +139,7 @@ explorer = {
         currency = ethereum_classic
         host = "https://explorers.api.vault.ledger.com"
         host = ${?WALLET_ETC_EXPLORER_ENDPOINT}
-        port = 443
+        port = 80
```
RomanWlm, line 142:
```suggestion
        port = 443
```

wd-local.env
```diff
@@ -0,0 +1,44 @@
...
+export WALLET_ETC_EXPLORER_VERSION=v3
+export WALLET_ETC_EXPLORER_PROXYUSE=false
+export FEES_ETC_PATH=/blockchain/v3/eth/fees
+
+
+
+
+
+
+
```
RomanWlm, line 44:
```suggestion
```

src/main/resources/application.conf.sample
```diff
@@ -121,42 +85,46 @@ explorer = {
...
         disable_sync_token = ${?DEFAULT_DISABLE_SYNC_TOKEN}
       }
       {
         currency = bitcoin
-        host = "https://api.ledgerwallet.com"
+        host = "https://explorers.api.vault.ledger.com"
         host = ${?WALLET_BTC_EXPLORER_ENDPOINT}
         port = 443
         port = ${?WALLET_BTC_EXPLORER_PORT}
         explorer_version = "v2"
```
RomanWlm, line 101:
```suggestion
        explorer_version = "v3"
```

dud225 (commented):
I fully support tthis endeavour :+1: 
I've 2 comments:
- We may need to configure the WD to point to a different explorer endpoint for all coins (eg to test a new infrastructure). There are a lot of coins, we may miss some of them. We would need to be able to overwrite the default settings.
- There may always be a gap between the coins defined within the WD and coins coming in on the Vault. Many coins have similarities with existings ones eg BTC, so it would be great to be able to dynamically add new coin settings in some way.

PR #270, opened 2020-09-15
ripple: add new endpoint to get hash of a signed transaction

onyb:
|JIRA|[LV-2543](https://ledgerhq.atlassian.net/browse/LV-2543)|
-|-

XRP is kind of a crazy beast. It is possible to have a transaction that failed during broadcast (ex, due to the `tecNO_DST_INSUF_XRP` error), but still ended up being recorded in the blockchain (because the protocol still wants to burn the fees).

This can cause inconsistencies during reconciliation on the Gate. The wallet daemon provides us with the transaction hash IFF the broadcast was successful. However, when it's a failure, we don't have the hash, and hence the Gate never knows about the transaction hash that was recorded on the blockchain.

The idea is to call the `/hash` endpoint to get the hash of the signed transaction. I'm limiting the scope of this PR to only Ripple. Need this for Vault 3.3.0 (upcoming release). For the full fix to work, the Gate also relies on https://github.com/LedgerHQ/ledger-wallet-daemon/commit/05f9f69da5e025a80e16967ca5c249c8c4265303.

Corresponding Gate PR: https://github.com/LedgerHQ/ledger-vault-api/pull/1345


src/main/scala/co/ledger/wallet/daemon/models/Account.scala
```diff
@@ -235,17 +238,34 @@ object Account extends Logging {
...
+      }
+    } yield hash
+  }
+
+  def broadcastXRPTransaction(rawTx: Array[Byte], signature: XRPSignature, a: core.Account, c: core.Currency)(implicit ec: ExecutionContext): Future[String] = {
+    for {
+      tx <- getSignedXRPTransaction(rawTx, signature, c)
+      txHash <- a.asRippleLikeAccount.broadcastTransaction(tx)
+    } yield txHash
+  }
```
vbergeron-ledger, line 267:
```suggestion
  def broadcastXRPTransaction(rawTx: Array[Byte], signature: XRPSignature, a: core.Account, c: core.Currency)(implicit ec: ExecutionContext): Future[String] = {
    getSignedXRPTransaction(rawTx, signature, c)
      .flatMap(a.asRippleLikeAccount.broadcastTransaction)
  }
```

src/main/scala/co/ledger/wallet/daemon/services/TransactionsService.scala
```diff
@@ -62,4 +62,17 @@ class TransactionsService @Inject()(defaultDaemonCache: DefaultDaemonCache, mess
...
+      case (account, wallet) =>
+        for {
+          r <- wallet.getWalletType match {
+            case WalletType.RIPPLE =>
+              val req = messageBodyManager.read[BroadcastTransactionRequest](request)
+              account.getXRPTransactionHash(req.hexTx, req.hexSig, wallet.getCurrency)
+            case w => Future.failed(CurrencyNotFoundException(w.name()))
+          }
+        } yield r
+    }
```
vbergeron-ledger, line 76:
```suggestion
      case (account, wallet) => wallet.getWalletType.flatMap {
            case WalletType.RIPPLE =>
              val req = messageBodyManager.read[BroadcastTransactionRequest](request)
              account.getXRPTransactionHash(req.hexTx, req.hexSig, wallet.getCurrency)
            case w => Future.failed(CurrencyNotFoundException(w.name()))
          }
    }
```

vbergeron-ledger (approved):

src/main/scala/co/ledger/wallet/daemon/models/Currency.scala
```diff
@@ -43,6 +44,12 @@ object Currency {
       case w => Left(s"$w is not Ripple")
     }
 
+  def parseSignedXRPTransaction(currency: core.Currency)(rawTx: Array[Byte]): Either[String, core.RippleLikeTransaction] =
+    currency.getWalletType match {
+      case core.WalletType.RIPPLE => Right(core.RippleLikeTransactionBuilder.parseRawSignedTransaction(currency, rawTx))
+      case w => Left(s"$w is not Ripple")
```
RomanWlm, line 50:
```suggestion
      case w => Left(s"Wrong currency provided $w is not Ripple")
```

RomanWlm (approved):

PR #276, opened 2020-10-06
publish latest known block height and push ERC20 balances on parent account synchronization

dkremer-ledger:


vbergeron-ledger (changes requested):
Small changing to be made to have a cleaner interface.

vbergeron-ledger (approved):

PR #277, opened 2020-10-09
BACK-831: Fix regression due to doubling of gas limit

onyb:
There are two situations when a raw transaction is built by the wallet daemon:
1. Computing fees.
2. Building the actual transaction to sign.

Recently, #272 (JIRA: BACK-831) was introduced to double the gas limit estimates provided by the explorers. However, this doubling should only be done when the Gate has not specified the precise gas limit to be used for transaction creation, otherwise, the gas limit will be quadrupled instead of doubled.

### DIFF
--- a/src/main/scala/co/ledger/wallet/daemon/models/Account.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/models/Account.scala
@@ -28,6 +28,7 @@ import com.twitter.inject.Logging
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.math.BigDecimal
 
 object Account extends Logging {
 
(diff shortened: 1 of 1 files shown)


YBadiss (approved):

src/main/scala/co/ledger/wallet/daemon/models/Account.scala
```diff
@@ -337,15 +338,20 @@ object Account extends Logging {
...
           }
+
           gasLimit <- ti.gasLimit match {
             case Some(amount) => Future.successful(amount)
             case None => ClientFactory.apiClient.getGasLimit(
-              c, ti.contract.getOrElse(ti.recipient), Some(erc20Account.getAddress), Some(inputData))
+                c, ti.contract.getOrElse(ti.recipient), Some(erc20Account.getAddress), Some(inputData)
+            ).map {
+              v => (BigDecimal(v) * BigDecimal(DaemonConfiguration.ETH_SMART_CONTRACT_GAS_LIMIT_FACTOR))
+                .setScale(0, BigDecimal.RoundingMode.CEILING).toBigInt()
```
vbergeron-ledger, line 348: Maybe put that in a function it will be clearer when reading.

vbergeron-ledger (approved):

PR #282, opened 2020-10-30
Back 1012 datadog agent with jdk8 upgrade

mbasire-ledger:
In order to monitor the behavior of the wallet daemon, the datadog jvm agent has been attached.


Dockerfile
```diff
@@ -9,11 +9,20 @@ ADD . /build
...
 #### RUN STEP ###
-FROM openjdk:8u181-jre-slim-stretch
+FROM openjdk:8u272-jre-slim-buster
 ENV HTTP_PORT 9200
 ENV ADMIN_PORT 0
 ENV STAGE dev
 
+## Datadog environment
+ENV DD_SERVICE=wallet-daemon
+ENV DD_VERSION=2.9.0-rc4
```
RomanWlm, line 19: We will need to automate this, otherwise we will miss to update it frequently 👀

RomanWlm (approved):

RomanWlm:
Goss validation is still failing, we should try to run locally docker images in order to access to logs

RomanWlm (changes requested):
HODL as goss validation is not working properly

Dockerfile
mbasire-ledger: DD_VERSION has been set to COMMIT_HASH.

By the way, DD_SERVICE should not be changed. What other value may it contain ?

Dockerfile
```diff
@@ -9,11 +11,26 @@ ADD . /build
...
+## Datadog environment
+ENV DD_SERVICE=wallet-daemon
+ENV DD_VERSION=$docker_tag
+ENV DD_ENV=STAGE
+
+ENV DD_AGENT_HOST=172.17.0.1
+ENV DD_TRACE_DEBUG=false
+ENV DD_INTEGRATIONS_ENABLED=false
+ENV DD_LOGS_INJECTION=false
+ENV DD_PROFILING_ENABLED=false
```
YBadiss, line 31: why do you disable them all?

mbasire-ledger: I disabled all of them because of image testing step in CI.
I was thinking that disabled by default was not a bad thing.
I hoped datadog would be activated at deployment. 
What do you think about this ?

Dockerfile
mbasire-ledger: DD_VERSION get the DOCKER_TAG at image building step. It can  be overriden at deployment time.

RomanWlm (approved):

vbergeron-ledger (approved):

mbasire-ledger:
It seems datadog has been integrated in another PR. Maybe we should close this one ?

PR #285, opened 2020-11-04
BACK-1079 AccountSynchronizer becomes an Actor

mbasire-ledger:
The AccountSynchronizer does not block threads anymore. 
It delegates operations publishing to another actor. 
Both AccountSynchronizer and AccountOperationsPublisher share the same wd-blocking dispatcher for libcore jobs.

### DIFF
--- a/src/main/scala/co/ledger/wallet/daemon/services/AccountSynchronizer.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/services/AccountSynchronizer.scala
@@ -1,31 +1,34 @@
 package co.ledger.wallet.daemon.services
 
 import java.util.Date
-import java.util.concurrent.atomic.{AtomicLong, AtomicReference}
-import java.util.concurrent.{ConcurrentHashMap, Executors, Semaphore}
+import java.util.concurrent.{ConcurrentHashMap, Executors}
 
+import akka.actor.Status.Failure
+import akka.actor.{Actor, ActorLogging, ActorRef, PoisonPill, Timers}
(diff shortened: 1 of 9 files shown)


src/main/scala/co/ledger/wallet/daemon/services/AccountSynchronizer.scala
```diff
@@ -151,7 +158,11 @@ class AccountSynchronizerManager @Inject()(daemonCache: DaemonCache, synchronize
 
   // return None if account info not found
   def getSyncStatus(accountInfo: AccountInfo): Option[SyncStatus] = {
-    Option(registeredAccounts.get(accountInfo)).map(_.getSyncStatus)
+    Option(registeredAccounts.get(accountInfo)).flatMap { synchronizer =>
+      implicit val timeout: Timeout = Timeout(10 seconds)
+      val status = ask(synchronizer, GetStatus).mapTo[SyncStatus]
+      Try(Await.result(status, 10 seconds)).toOption
```
RomanWlm, line 164: Finally we use Await here ?

RomanWlm (changes requested):
HODL for CI pass

src/main/scala/co/ledger/wallet/daemon/services/AccountSynchronizer.scala
mbasire-ledger: This method was already awaiting for a result. We should change the result to a Future of something in a next release.

RomanWlm (approved):

PR #293, opened 2020-11-18
Fix get status endpoint

mbasire-ledger:
Get status endpoint does not block anymore

The account synchronizer in charge of giving the status used a dispatcher busy by active sync.
It uses now the default dispatcher.


RomanWlm (approved):

RomanWlm (changes requested):
Could we, in case of Await timeout, return 500 error ? 404 is not accurate and we are falling down inside the same workflow than we have when account does not exists

RomanWlm (approved):

PR #297, opened 2020-11-18
BACK 1119 Return to idle state on failure and schedule a sync 

mbasire-ledger:
The account synchronizer replies only on ForceSynchronization and GetStatus.
All known messages are handled in each state.

On error: return to idle state and schedule a sync


src/main/scala/co/ledger/wallet/daemon/services/AccountSynchronizer.scala
```diff
@@ -242,8 +242,9 @@ class AccountSynchronizer(account: Account,
   def uninitialized() : Receive = {
     case Init => lastAccountBlockHeight.pipeTo(self)
     case h: BlockHeight => context become idle(lastHeightSeen = h)
-      schedulePeriodSync()
+      scheduleNextSync()
 
+    case StartSynchronization =>
```
RomanWlm, line 247: Add comment to explain that we do not expect any processing

src/main/scala/co/ledger/wallet/daemon/services/AccountSynchronizer.scala
```diff
@@ -269,46 +270,53 @@ class AccountSynchronizer(account: Account,
...
-        .andThen{ case scala.util.Failure(_) => context become idle(lastHeightSeen) }
 
-    case Failure(t) => log.error(t, s"An error occurred synchro synchronizing account $accountInfo  at height $lastHeightSeen")
+    case Failure(t) => log.error(t, s"An error occurred synchronizing account $accountInfo  at $lastHeightSeen")
+      scheduleNextSync()
   }
 
   private def synchronizing(fromHeight: BlockHeight) : Receive = {
-    case GetStatus | ForceSynchronization | StartSynchronization => updatedSyncingStatus(fromHeight).pipeTo(sender())
+    case StartSynchronization =>
```
RomanWlm, line 279: Same here :)

RomanWlm (changes requested):
Few comments inline. 
As I understand we need now to reschedule manually the next sync. 
Let's discuss about the PR together 👍

mbasire-ledger:
> Few comments inline.
> As I understand we need now to reschedule manually the next sync.
> Let's discuss about the PR together

This is not really mandatory but it is safe to reschedule. The eventual previous scheduled StartSynchronization is cancelled. 
StartSynchronization was still unscheduled while handling ReSync command (wiping account operations data).

Let me remind you that the timer's delay is actually short by default. ie, after the end of a sync or a failure, the AccountSynchronizer will wait $SYNC_INTERVAL_IN_SECONDS (default : 1 minute) before retry a sync.

It means is **not** really **periodic** but automatic. The real problem : an account synchronizer can not be stopped. It has to be killed by its manager (a future feature maybe ?).

RomanWlm (approved):
Pon Poulot

PR #304, opened 2020-11-30
Back 1206 custom gap limit wallet config for sync

RomanWlm:


YBadiss (approved):

RomanWlm:
I close it as this is not necessary thanks to get fresh addresses servce

PR #306, opened 2020-12-01
Fix force synchro return status

mbasire-ledger:
The wallet daemon was not able to return a status even if it succeeded to trigger a sync


RomanWlm:
Thanks for this PR, this bug has been addressed since then.

PR #311, opened 2020-12-05
Back 1219 pool dto caching

RomanWlm:
PoolDto are immutable and requested for every requests received.
There is only few lines to cache (one per workspaces)
As recent trouble we had it make sens to put them behind caching.


RomanWlm:
Has been addressed on 2.8.x

PR #318, opened 2020-12-10
Update all XTZ code to latest NRT changes

twilgenbus-ledger:
This PR tries to cherry pick all work from the following PRs:
- #279 
- #316


docker-compose.yml
```diff
@@ -15,23 +15,14 @@ services:
...
-      CORE_PG_PWD: ledger_secret
-      CORE_PG_DB_NAME_PREFIX: wallet_daemon_
-    entrypoint: /app/run.sh -Dscala.concurrent.context.numThreads=x8 -Dscala.concurrent.context.maxThreads=x8
-    healthcheck:
-      test: "command -v curl && curl -fsS http://wallet-daemon:9200/_health"
-      interval: 30s
-      timeout: 5s
-      retries: 3
-      start_period: 30s
+  rabbitmq:
```
RomanWlm, line 18: Are theses changes intended to be included ? 
Can you motivate them ?

src/it/scala/co/ledger/wallet/daemon/api/XTZAccountsApiTest.scala
```diff
@@ -0,0 +1,182 @@
...
+import com.twitter.finagle.http.Status
+
+class XTZAccountsApiTest extends APIFeatureTest {
+  val poolName = "tez_test_pool"
+
+  override def beforeAll(): Unit = {
+    createPool(poolName)
+    // when(publisher.publishAccount(any[Account], any[Wallet], meq(poolName), any[SyncStatus])).thenReturn(Future.unit)
+    // when(publisher.publishOperation(any[OperationView], any[Account], any[Wallet], meq(poolName))).thenReturn(Future.unit)
+  }
```
RomanWlm, line 19:
```suggestion
```

src/it/scala/co/ledger/wallet/daemon/api/XTZAccountsApiTest.scala
```diff
@@ -0,0 +1,182 @@
...
+  test("Create XTZ account") {
+    val walletName = "xtzWalletAccountCreation"
+    assertWalletCreation(poolName, walletName, "tezos", Status.Ok)
+    assertCreateAccount(CORRECT_BODY_XTZ, poolName, walletName, Status.Ok)
+    val addresses = parse[Seq[FreshAddressView]](assertGetFreshAddresses(poolName, walletName, index = 0, Status.Ok))
+    assert(addresses.nonEmpty)
+    info(s"Here are addresses : $addresses")
+    assertSyncAccount(poolName, walletName, 0)
+    val operations = parse[Map[String, JsonNode]](assertGetAccountOps(poolName, walletName, 0, OperationQueryParams(None, None, 1000, 0), Status.Ok))
+    assert(operations.nonEmpty)
```
RomanWlm, line 47: We could probably enhance checks here, for instance check that very firsts operations of the account are effectively retrieved.

src/it/scala/co/ledger/wallet/daemon/api/XTZAccountsApiTest.scala
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  test("Create XTZ Transaction") {
```
RomanWlm, line 50: This function could be split in several ones : 
- One for No fees provided => BadRequest
- One for Not enought funds case
- Third one for Fees level check

src/it/scala/co/ledger/wallet/daemon/api/XTZAccountsApiTest.scala
```diff
@@ -0,0 +1,182 @@
...
+    val addresses = parse[Seq[FreshAddressView]](assertGetFreshAddresses(poolName, walletName, index = 0, Status.Ok))
+    assert(addresses.nonEmpty)
+    assertSyncAccount(poolName, walletName, 0)
+    val operations = parse[Map[String, JsonNode]](assertGetAccountOps(poolName, walletName, 0, OperationQueryParams(None, None, 1000, 0), Status.Ok))
+    assert(operations.nonEmpty)
+    // Create transaction to self
+    val txNormalSpeedRequest = CreateXTZTransactionRequest(TezosOperationTag.OPERATION_TAG_TRANSACTION, addresses.head.address, "100", false, Some("800"), Some("800"), Some("322"), Some("NORMAL"))
+
+    val txNormalSpeedRequestJson = server.mapper.writeValueAsString(txNormalSpeedRequest)
+    // FIXME: Should be Status.BadRequest
```
RomanWlm, line 176: This should be fixed by having appropriate case on ExceptionMapper. Let me know If I can help on this

src/it/scala/co/ledger/wallet/daemon/api/XTZAccountsApiTest.scala
```diff
@@ -0,0 +1,182 @@
...
+import co.ledger.core.TezosOperationTag
+import co.ledger.wallet.daemon.controllers.TransactionsController.CreateXTZTransactionRequest
+import co.ledger.wallet.daemon.models.coins.UnsignedTezosTransactionView
+import co.ledger.wallet.daemon.models.{DelegationView, FreshAddressView}
+import co.ledger.wallet.daemon.services.OperationQueryParams
+import co.ledger.wallet.daemon.utils.APIFeatureTest
+import com.fasterxml.jackson.databind.JsonNode
+import com.twitter.finagle.http.Status
+
+class XTZAccountsApiTest extends APIFeatureTest {
```
RomanWlm, line 12: I would suggest to add empty account creation which is an interesting edge case IMO

src/main/resources/application.conf.sample
```diff
@@ -91,7 +91,7 @@ explorer = {
         currency = default
         host = "https://explorers.api.vault.ledger.com"
         port = 443
-        proxyuse = true
+        proxyuse = false
```
RomanWlm, line 94: proxyuse change are ok as now we are not behind proxy since EKS migration but I would suggest to let them in a separated commit (please keep this in mind when you will squash commits before merge).
Usefull for easy revert and separate purpose.

src/main/resources/application.conf.sample
```diff
@@ -438,6 +450,15 @@ ethereum = {
   gaslimitfactor = ${?ETH_GAS_LIMIT_FACTOR}
 }
 
+tezos = {
+node = {
```
RomanWlm, line 454: Just wondering if we could add it easily into the coin config directly 🤔 
I think we are good as it for now 👍

src/main/scala/co/ledger/wallet/daemon/clients/ApiClient.scala
```diff
@@ -189,4 +189,11 @@ object ApiClient {
     }
   }
 
+  case class XtzFeeInfo(minimumFee: BigInt) extends FeeInfo {
+    def getAmount(feeMethod: FeeMethod): BigInt = feeMethod match {
+      case FeeMethod.SLOW => minimumFee
+      case FeeMethod.NORMAL => minimumFee * 2
+      case FeeMethod.FAST => minimumFee * 10
```
RomanWlm, line 196: Are theses Fees policy specified somewhere ? 
Would we want to let us ability to tune them later ? (I mean expose them into ENV config as for ETh ones)
Open question.

src/main/scala/co/ledger/wallet/daemon/configurations/DaemonConfiguration.scala
```diff
@@ -208,7 +208,7 @@ object DaemonConfiguration extends Logging {
       val port = path.getInt("port")
       val url = new URL(s"$host:$port")
       // Use proxy if proxy is enabled globally and proxyuse is true or undefined
-      val proxyUse = Try(path.getBoolean("proxyuse")).getOrElse(true)
+      val proxyUse = Try(path.getBoolean("proxyuse")).getOrElse(false)
```
RomanWlm, line 211: 👍 keep it in same commit as proxyuse change if this is possible

src/main/scala/co/ledger/wallet/daemon/models/Account.scala
```diff
@@ -601,6 +665,18 @@ object Account extends Logging {
...
   }
+
+  def getDelegation(a: Account)(implicit ec: ExecutionContext): Future[Seq[DelegationView]] = {
+    a.getWalletType match {
+      case WalletType.TEZOS => a.asTezosLikeAccount().getCurrentDelegate()
+        .filter(address => !address.isEmpty)
+        .map(DelegationView.fromDelegatedAddress)
+        .map(List(_))
+        .recover { case _: java.util.NoSuchElementException => List[DelegationView]() }
+      case _ => Future.successful(List())
```
RomanWlm, line 676: Do we really want to return gracefully when the delegation is called on an unsupported coin operation ?

src/main/scala/co/ledger/wallet/daemon/models/Account.scala
```diff
@@ -601,6 +665,18 @@ object Account extends Logging {
...
     new Derivation(coreD)
   }
+
+  def getDelegation(a: Account)(implicit ec: ExecutionContext): Future[Seq[DelegationView]] = {
+    a.getWalletType match {
+      case WalletType.TEZOS => a.asTezosLikeAccount().getCurrentDelegate()
+        .filter(address => !address.isEmpty)
+        .map(DelegationView.fromDelegatedAddress)
+        .map(List(_))
+        .recover { case _: java.util.NoSuchElementException => List[DelegationView]() }
```
RomanWlm, line 675: Few questions 
- Probably coin specific but when do we expect empty address and what does it mean ?
- When `java.util.NoSuchElementException` can be thrown ? Is empty list expected here ? We could probably at least log the error and let future be failure. What do you think ?

src/main/scala/co/ledger/wallet/daemon/models/Pool.scala
```diff
@@ -234,6 +236,12 @@ class Pool(private val coreP: core.WalletPool, val id: Long) extends Logging {
       if (currencyName == "ripple") {
         walletConfig.putString("BLOCKCHAIN_EXPLORER_API_ENDPOINT", s"${apiUrl.getProtocol}://${apiUrl.getHost}")
         walletConfig.putString("BLOCKCHAIN_EXPLORER_PORT", apiUrl.getPort.toString)
+      // FIXME get help about having a path already inside the endpoint
+      // it creates a wrong URL if not hardcoded, like https://xtz-explorer.api.live.ledger.com:-1/explorer:443
+      // which doesn't work because Core explorer HTTP calls end up being
+      // GET https://xtz-explorer.api.live.ledger.com:-1/explorer:443/block/head
```
RomanWlm, line 242: What happens if we do not provide port number ? 
IMO we can leave with it... 
I suggest to keep s"${apiUrl.getProtocol}://${apiUrl.getHost}/explorer" or something similar to keep endpoint configurable.
Let's raise a libcore ticket about this issue

src/main/scala/co/ledger/wallet/daemon/models/coins/Tezos.scala
```diff
@@ -0,0 +1,172 @@
...
+      Option(from.getGasLimit).map(limit => limit.toBigInt.asScala),
+      Try(from.getStorageLimit).toOption.map(limit => limit.asScala),
+      from.getBlockHash,
+      from.getStatus
+    )
+  }
+
+  // private def newBlockView(from: core.TezosLikeBlock): BlockView = {
+  //   CommonBlockView(from.getHash, from.getHeight, from.getTime)
+  // }
```
RomanWlm, line 51: If we do not need it we can just remove this part

tooling/database/postgres/dbs.sh
```diff
@@ -24,7 +24,7 @@ $command -h $host -p $port wallet_daemon -U $user
 echo "     |--> wallet_daemon Done"
 
 # Databases used in IT tests
-tests=(default_test_pool create_delete_pool walletpooldao databasedaotest account_test pool_a pool_b pool_c empty_pool)
+tests=(default_test_pool create_delete_pool walletpooldao databasedaotest account_test pool_a pool_b pool_c empty_pool test_pool random POOL_NAME tez_test_pool)
```
RomanWlm, line 27:
```suggestion
tests=(default_test_pool create_delete_pool walletpooldao databasedaotest account_test pool_a pool_b pool_c empty_pool tez_test_pool)
```

RomanWlm (changes requested):
Thank you for your Pull request ! Tezos implem has been well addressed 🚀 ! 
Few comments bellow, I'am available to discuss to clarify if needed some of them 👍

tooling/database/postgres/dbs.sh
twilgenbus-ledger: Good catch! Probably an issue during merges.

src/main/scala/co/ledger/wallet/daemon/models/Account.scala
twilgenbus-ledger: > Probably coin specific but when do we expect empty address and what does it mean ?

In Tezos you can have either one or no action delegation. The address is empty if the latest is true.

> When java.util.NoSuchElementException can be thrown ?

I'm using `filter` to check if the aforementioned address is prevent. In case of a negative predicate, the future will then fail with `NoSuchElementException` (=> [relevant doc](https://www.scala-lang.org/api/2.12.3/scala/concurrent/Future.html#filter(p:T=%3EBoolean)(implicitexecutor:scala.concurrent.ExecutionContext):scala.concurrent.Future[T])). That means we end up using exception catching in a non-exceptional case. Maybe there's a more idiomatic way to end up with the same behaviour?

> Is empty list expected here ?

`getCurrentDelegate` does not return a list but a single address, if there's any.

> We could probably at least log the error and let future be failure.

This is not an exceptional behaviour per-se, we just end up having to catch this exception because this is how behave `Future.filter`.

src/main/scala/co/ledger/wallet/daemon/models/Account.scala
twilgenbus-ledger: I'd rather not throw an error in this case and felt like an empty list did the job.
I understand that it could be considered confusing, what kind of error would you think is appropriate?

docker-compose.yml
twilgenbus-ledger: Nope, I reverted it all. Let me know if that's resolved!

src/main/scala/co/ledger/wallet/daemon/clients/ApiClient.scala
twilgenbus-ledger: From the [Acceptance Plan](https://ledgerhq.atlassian.net/wiki/spaces/LCH/pages/1762525197/XTZ-LV+-+Acceptance+Plan):
```
Pending Q: Can we follow Vault fee management for Tezos network fees?
 - Slow: Current state - 25%
 - Medium: Current state of the network at broadcast (by default)
 - Fast: Current state + 25%
```
It was added by @gagbo. It seems to me it might have been a copy/paste from how XLM works. Maybe you could confirm?
Should we change it to what's in the acceptance plan?

twilgenbus-ledger: I've changed the implementation to what was in the acceptance plan.

PR #320, opened 2020-12-10
Move back to stable its - chunck 1

RomanWlm:
This PR is reducing the number of pool (then dbs) created to run you test harness
 It also fix missing configuration keys for its (since dd integration)


src/it/resources/application.conf
```diff
@@ -531,3 +531,18 @@ synchronization {
     }
 }
 
+datadog {
```
twilgenbus-ledger, line 534: Although the key is required for the tests to start. I don't know if we actually want to enable it? Especially since it might make the IT depend on datadog being up at the time.

src/it/scala/co/ledger/wallet/daemon/api/WalletsApiTest.scala
```diff
@@ -7,96 +7,102 @@ import co.ledger.wallet.daemon.utils.APIFeatureTest
 import com.twitter.finagle.http.{Response, Status}
 
 class WalletsApiTest extends APIFeatureTest {
+  private val poolName = "default_test_pool"
 
   test("WalletsApi#Create then Get same wallet from pool Return OK") {
-    createPool(WALLET_POOL)
-    val createdW = walletFromResponse(assertWalletCreation(WALLET_POOL, "my_wallet", "bitcoin", Status.Ok))
+    createPool(poolName)
```
twilgenbus-ledger, line 13: Since we now have only one pool for all the tests in this class, we might want to move `createPool` + `deletePool` to `before/after`?

twilgenbus-ledger (approved):
Much cleaner!

Only have two tiny non-blocking questions.

mbasire-ledger (approved):

src/it/resources/application.conf
RomanWlm: When `agent_host` is empty, the agent is disabled actually

src/it/scala/co/ledger/wallet/daemon/api/WalletsApiTest.scala
RomanWlm: absolutely 👍

PR #321, opened 2020-12-11
Back 1219 poll dto caching patch2.8.1 h3

RomanWlm:


src/it/scala/co/ledger/wallet/daemon/api/AccountsApiTest.scala
```diff
@@ -181,7 +181,7 @@ class AccountsApiTest extends APIFeatureTest {
     assert(walPoolView.findValue("wallet_count").asInt() == 3)
     info(s"Pool is = $pool and $walPoolView")
     deletePool(poolName)
-    getPool(poolName, Status.NotFound)
+    getPool(poolName, Status.BadRequest)
```
mbasire-ledger, line 184: Do the endpoint change ?

tooling/database/postgres/dbs.sh
```diff
@@ -22,7 +22,7 @@ $command -h $host -p $port wallet_daemon -w -U $user
 echo "     |--> wallet_daemon Done"
 
 # Databases used in IT tests
-tests=(transactionsExcludeUTXOTest transactionsPartialTest currency_pool test_pool test_pool1 op_pool op_pool_mal info_pool exist_pool fresh_addresses_pool list_pool account_pool delete_pool delete_account balance_pool pool_1 pool_2 pool_3 random_pool multi_pool_mal transactionsCreation4Test ledger anotha_pool same_pool this_pool my_pool your_pool wallet_pool duplicate_pool empty_pool multi_pool random POOL_NAME myPool test_wallet getUtxo_pool )
+tests=(transactionsExcludeUTXOTest transactionsPartialTest currency_pool test_pool test_pool1 op_pool op_pool_mal info_pool exist_pool fresh_addresses_pool list_pool account_pool delete_pool delete_account balance_pool pool_1 pool_2 pool_3 random_pool multi_pool_mal transactionsCreation4Test ledger anotha_pool same_pool this_pool my_pool your_pool wallet_pool duplicate_pool empty_pool multi_pool random POOL_NAME myPool test_wallet getUtxo_pool pool_name )
```
mbasire-ledger, line 25: you added a pool "pool_name" and there is still "POOL_NAME" but in some test cases, _POOL_NAME_ became _pool_name_, is that a mistake ?

mbasire-ledger (approved):

src/it/scala/co/ledger/wallet/daemon/api/AccountsApiTest.scala
RomanWlm: Yes, we return bad request now when we hit an non existing endoint

tooling/database/postgres/dbs.sh
RomanWlm: Something to remove, actually the endpoint that delete pool is loosing future that asking for pool deletion. So we return before the pool is effectively deleted. Indeed POOL_NAME and pool_name is same for PG. (was an deseperate attemps after having found "database wdtest_pool_name does not exists" I remove POOL_NAME.

PR #326, opened 2020-12-31
BACK-1318/BACK-1316 - user pref redis support (Backport 2.8.x)

RomanWlm:


docker-compose.yml
```diff
@@ -17,13 +21,16 @@ services:
       - "5432:5432"
     command: -c "ssl=on" -c "ssl_key_file=/etc/ssl/private/ssl-cert-snakeoil.key" -c "ssl_cert_file=/etc/ssl/certs/ssl-cert-snakeoil.pem"
   wallet-daemon:
-    image: ledgerhq/ledger-wallet-daemon:latest
+    build: .
```
mbasire-ledger, line 24: Good choice

src/main/scala/co/ledger/wallet/daemon/database/RedisPreferenceBackend.scala
```diff
@@ -0,0 +1,133 @@
...
+  type KeyPref = ByteBuffer
+  type ValuePref = ByteBuffer
+  val prefCache: LoadingCache[KeyPref, Option[ValuePref]] = CacheBuilder.newBuilder()
+    .maximumSize(2000)
+    .build[KeyPref, Option[ValuePref]](new CacheLoader[KeyPref, Option[ValuePref]] {
+      override def load(key: KeyPref): Option[ValuePref] = {
+        loadPrefEntry(key.array()).map[ValuePref](ByteBuffer.wrap)
+      }
+    })
+  private implicit val akkaSystem: ActorSystem = akka.actor.ActorSystem()
```
mbasire-ledger, line 42: It should be injected. don't you think ?

src/main/scala/co/ledger/wallet/daemon/database/RedisPreferenceBackend.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  private val redis: RedisClient = RedisClient(
+    conf.host, conf.port, conf.password, conf.db, conf.name, conf.connectionTimeout
+  )
+
+  def init(): Unit = {
+  }
+
+  override def get(bytes: Array[Byte]): Array[Byte] = {
+    prefCache.get(ByteBuffer.wrap(bytes)).map(_.array).orNull
```
mbasire-ledger, line 63: Do libcore expect a null when nothing has been found ? The `get()` signature let me think not.

mbasire-ledger (changes requested):
https://github.com/debasishg/scala-redis gives a scala redis client **sync** which should suit our needs. 
The PreferencesBackend expect **sync** responses.

src/main/scala/co/ledger/wallet/daemon/database/RedisPreferenceBackend.scala
twilgenbus-ledger: This line was directly copied from the [`PostgresPreferenceBackend`](https://github.com/LedgerHQ/ledger-wallet-daemon/blob/f39402c03c9c91247e3ae15549ff6915ca11b490/src/main/scala/co/ledger/wallet/daemon/database/PostgresPreferenceBackend.scala#L45).

twilgenbus-ledger:
@mbasire-ledger It has been re-implemented with the proposed library.

src/main/scala/co/ledger/wallet/daemon/models/Pool.scala
```diff
@@ -262,43 +262,62 @@ class Pool(private val coreP: core.WalletPool, val id: Long) extends Logging {
...
+    Try(config.getString("core.core_database_engine")).toOption.getOrElse("sqlite3") match {
       case "postgres" =>
         info("Using PostgreSql as core database engine")
         val dbName = for {
           dbPort <- Try(config.getString("postgres.port"))
           dbHost <- Try(config.getString("postgres.host"))
           dbUserName <- Try(config.getString("postgres.username"))
           dbPwd <- Try(config.getString("postgres.password"))
-          dbPrefix <- Try(config.getString("postgres.db_name_prefix"))
+          dbPrefix <- Try(config.getString("postgres.core.db_name_prefix"))
```
mbasire-ledger, line 307: This is not the good key on 2.8

twilgenbus-ledger: Nice catch!

mbasire-ledger (approved):

src/main/resources/application.conf.sample
```diff
@@ -17,6 +18,15 @@ sqlite3 = {
...
 }
 
+redis = {
+    name = "redis-wallet-daemon"
+    name = ${?REDIS_NAME}
+    host = "localhost"
+    host = ${?REDIS_HOST}
+    port = 6379
+    port = ${?REDIS_PORT}
+}
```
mbasire-ledger, line 28: where are the following configuration keys with their eventual env variables : 
 - redis.password
 - redis.db
 - redis.connection_timeout
?

The code redis backend implementation references those variables. By the way, there is no redis.name references in the preferences backend.

twilgenbus-ledger:
```
redis.password
redis.db
redis.connection_timeout
```
Are all optional (see the pattern `Try().toOption` in the class `Pool`

Good catch on the redis.name. This is an artefact that should have been removed.

src/it/resources/application.conf
```diff
@@ -17,6 +17,16 @@ core = {
   http_client_retries = ${?CORE_HTTP_RETRIES}
 }
 
+redis = {
+    host = "localhost"
+    port = 6379
+}
+
+preferences {
```
RomanWlm, line 25: `user-preferences` is probably a bit more accurate

src/main/resources/application.conf.sample
RomanWlm: So, In order to be deployed and plugged on production existing redis instance, we would need to expose them into `application.conf.sample` with corresponding environment variables 
ie something like  : 
```
redis {
    host = "localhost"
    host = ${?REDIS_HOST}
    port = 6379
    port = ${?REDIS_PORT}
    password = ledger_secret
    password = ${REDIS_PASSWORD}
    db = default (?)
    db = ${REDIS_DB_NAME}
    connection_timeout = 2000 #ms ?
    connection_timeout = ${REDIS_CNX_TIMEOUT}
}
```

src/main/resources/application.conf.sample
```diff
@@ -17,6 +18,18 @@ sqlite3 = {
...
 }
 
+redis = {
+    host = "localhost"
+    host = ${?REDIS_HOST}
+    port = 6379
+    port = ${?REDIS_PORT}
+}
+
+preferences {
```
RomanWlm, line 28: `user-preferences` is probably a bit more accurate
also what are accepted/possible values for this one (any other value will let us fallback on levelDb ?

src/main/scala/co/ledger/wallet/daemon/models/Pool.scala
```diff
@@ -262,16 +262,21 @@ class Pool(private val coreP: core.WalletPool, val id: Long) extends Logging {
...
 
   def newCoreInstance(poolDto: PoolDto): Future[core.WalletPool] = {
-    val builder = core.WalletPoolBuilder.createInstance()
     val poolConfig = core.DynamicObject.newInstance()
-    Try(config.getString("core_database_engine")).toOption.getOrElse("sqlite3") match {
+    val builder = core.WalletPoolBuilder.createInstance()
+
+    config.getString("preferences.storage") match {
+      case "redis" => configureRedisAsPreferenceStorage(poolDto, builder)
+      // else let libcore use leveldb
```
RomanWlm, line 276: I think this is more safe to accept redis or leveldb but fail on any other values (avoid misconfiguration issues that will pass the deployment phase)

RomanWlm: Also, could we add a little Info log for preference storage configured on levelDb ? (The existing one for redis is already existing 👍 )

src/main/scala/co/ledger/wallet/daemon/models/Pool.scala
```diff
@@ -319,6 +318,33 @@ object Pool extends Logging {
...
+        port = port,
+        password = Try(config.getString("redis.password")).toOption,
+        db = Try(config.getInt("redis.db")).toOption,
+        connectionTimeout = Try(config.getInt("redis.connection_timeout")).toOption
+      )
+    }
+
+    maybeRedisConf match {
+      case Success(redisConfig) =>
+        info(s"Using Redis as core preference database $redisConfig")
```
RomanWlm, line 338: Avoid to log redis db password :)

RomanWlm: copy(password = "****") can be fine here

src/main/resources/application.conf.sample
twilgenbus-ledger: Agreed, a couple more precisions first:
- db: it's an integer, useful only in cluster mode (useful only if we were reach huge numbers of preferences and require >20Go) => it will most likely never be changed
- password: usage of a password in Redis is recommended only if the instance will be exposed to the internet (which is not what it was designed for, [more](https://redis.io/topics/security)) => it will most likely never be changed
- connection_timeout: redis just doesn't close clients connections by default for performance reasons. This is mostly useful in weird clients configuration (i.e: a rogue number of clients and/or connections) or a debug situation where a controlled client would have a bug. [more](https://redis.io/topics/clients) => it will most likely never be changed

I believe it would be better to just have the lines:
```
password = ${?REDIS_PASSWORD}
db = ${?REDIS_DB_NAME}
connection_timeout = ${?REDIS_CONNECTION_TIMEOUT}
```
And no defaults.

src/main/scala/co/ledger/wallet/daemon/models/Pool.scala
twilgenbus-ledger: Agreed on both ideas 👍

src/main/scala/co/ledger/wallet/daemon/models/Pool.scala
twilgenbus-ledger: Good catch!

src/main/resources/application.conf.sample
```diff
@@ -17,6 +18,23 @@ sqlite3 = {
...
 
+redis = {
+    host = "localhost"
+    host = ${?REDIS_HOST}
+    port = 6379
+    port = ${?REDIS_PORT}
+    password = ${?REDIS_PASSWORD}
+    db = 0
+    db = ${?REDIS_DB_NUMBER}
+    connection_timeout = 2000
```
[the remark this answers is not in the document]
twilgenbus-ledger: Yes 👍

PR #332, opened 2021-01-29
Back 1218 add datadog support 2.8.x

RomanWlm:
BACK-1218 - Add datadog agent for 2.8.x

    * add agent with stb agent plugin
    * run.sh is passing datadog agent variable only when DATADOG HOST is declared
    * Do not enb"ed datadog agent when DD agent host is declared (native
    packager agent script)
    * Log proxy and datadog whebn enabled on run.sh


Dockerfile
```diff
@@ -9,11 +9,16 @@ ADD . /build
 RUN ./docker/build.sh
 
 #### RUN STEP ###
-FROM openjdk:8u181-jre-slim-stretch
+FROM openjdk:8u272-jre-slim-buster
+ARG docker_tag
```
mbasire-ledger, line 13: do the CI gives the value during image building ?

mbasire-ledger (approved):

Dockerfile
RomanWlm: You're right, this is missing.
I'll update this based on your work on that

mbasire-ledger: Perfect

PR #334, opened 2021-02-05
[BACK-1486] Make the logs great again

twilgenbus-ledger:
Inspired by [this commit](https://github.com/LedgerHQ/ledger-wallet-daemon/commit/98b0d0a187a9151e40cff867093e3d6aaf390746).

It takes an even more radical stand (info changed to warn), proposition is that we will come back to it if we need some special debug. But at least, we might be able to convince intra to send all the logs to the datadogs.

### DIFF
--- a/src/it/resources/logback-test.xml
+++ b/src/it/resources/logback-test.xml
@@ -4,13 +4,22 @@
             <pattern>%date %thread TraceId=%X{traceId} [%level] %logger{0} : %msg%n</pattern>
         </encoder>
     </appender>
-    <logger name="co.ledger.wallet.daemon" level="debug" />
-    <logger name="co.ledger.wallet.daemon.database.core" level="error" />
-    <logger name="slick" level="error" />
-    <logger name="io.netty" level="error" />
-    <logger name="com.twitter" level="error" />
-    <logger name="djinni" level="error" />
-    <root level="debug">
-        <appender-ref ref="STDOUT-test" />
+    <variable name="LOG_LEVEL" value="${LOG_LEVEL:-WARN}" />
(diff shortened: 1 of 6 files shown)


krauscher-ledger:
My 2 cents on that is that log level should be a runtime configuration.

krauscher-ledger:
Note that you can use environment variables in logback configuration. So you could have an environment variable for each logger level (with a default value). This prevent you to have to redeploy a new version each time you need more log.

krauscher-ledger:
Syntax is 
```level="${LOG_LEVEL:-INFO}"```

RomanWlm (approved):
Great !

RomanWlm:
Few remarks : Utils.preview is limiting ouput pauload logging to 200 chars today, we can enhance this value to get more info on returned payoads

RomanWlm:
> My 2 cents on that is that log level should be a runtime configuration. (and not hardcoded)

Totally agree, furthermore having good default values is also great things to have

twilgenbus-ledger:
> * Utils.preview is limiting ouput payload logging to 200 chars today, we can enhance this value to get more info on returned payoads
- Good catch, will upgrade to 1000
> * Does libcore reduced logs when synchronizing ? (If I remember it's about 6 to 10 lines of logs per explorer page retrievals)
- Good point, no way to control that from the WD side from what I understand. We should probably raise a ticket to the libcore team.

twilgenbus-ledger:
@krauscher-ledger excellent idea! I'll add variables for each entry. We'll still have to add new entries/variables within logback during a exploratory debug step, but at least we won't have to redeploy during existing log levels fine tunings.

krauscher-ledger (approved):
lgtm

PR #336, opened 2021-02-12
BACK-1471 - Expose endpoint to force create ER20 accounts

nukky94:
Expose an endpoint to take an ERC20 contract addresses list and call the newly-created libcore endpoint to force ERC20 accounts creation in libcore DB


mbasire-ledger (approved):
We will improve ETH test coverage later when the POC is approved.

PR #343, opened 2021-02-22
[BACK-1565] Add implementation of fetching ERC20 operations from last block height

twilgenbus-ledger:
This PR solves BACK-1565 wheer every single operations were pushed instead of all operations from the given block height.

It does so through the usage of WalletPoolDao, to continue the replacement of going through Libcore's dedicated system.

### DIFF
--- a/src/main/scala/co/ledger/wallet/daemon/database/core/WalletPoolDao.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/database/core/WalletPoolDao.scala
@@ -67,6 +67,12 @@ class WalletPoolDao(poolName: String)(implicit val ec: ExecutionContext) extends
   override def findERC20OperationsByUids(a: Account, w: Wallet, filteredUids: Seq[ERC20OperationUid], offset: Int, limit: Int): Future[Seq[OperationView]] =
     erc20daoForWalletType(w.getWalletType: WalletType).findERC20OperationsByUids(a, w, filteredUids, offset, limit)
 
+  /**
+    * List erc20 operations from an account starting at specified bock height
+    */
+  override def findERC20OperationsFromBlockHeight(a: Account, w: Wallet, blockHeight: Long, offset: Int, limit: Int): Future[Seq[OperationView]] =
+    erc20daoForWalletType(w.getWalletType: WalletType).findERC20OperationsFromBlockHeight(a, w, blockHeight, offset, limit)
+
 }
 
 object WalletPoolDao {
--- a/src/main/scala/co/ledger/wallet/daemon/database/core/WalletPoolDao.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/database/core/WalletPoolDao.scala
@@ -68,7 +68,7 @@ class WalletPoolDao(poolName: String)(implicit val ec: ExecutionContext) extends
     erc20daoForWalletType(w.getWalletType: WalletType).findERC20OperationsByUids(a, w, filteredUids, offset, limit)
 
   /**
-    * List erc20 operations from an account starting at specified bock height
+    * List erc20 operations from an account starting at specified block height
     */
   override def findERC20OperationsFromBlockHeight(a: Account, w: Wallet, blockHeight: Long, offset: Int, limit: Int): Future[Seq[OperationView]] =
     erc20daoForWalletType(w.getWalletType: WalletType).findERC20OperationsFromBlockHeight(a, w, blockHeight, offset, limit)
--- a/src/main/scala/co/ledger/wallet/daemon/database/core/WalletPoolDao.scala
+++ b/src/main/scala/co/ledger/wallet/daemon/database/core/WalletPoolDao.scala
@@ -13,7 +13,7 @@ class WalletPoolDao(poolName: String)(implicit val ec: ExecutionContext) extends
   val db: Database = new Database(DaemonConfiguration.coreDbConfig, poolName)
 
   lazy val btcDao = new BitcoinDao(db)
-  lazy val ethDao = new EthereumDao(db)
+  lazy val ethDao = new EthereumDao(db, poolName)
   lazy val xrpDao = new RippleDao(db)
   lazy val xlmDao = new StellarDao(db)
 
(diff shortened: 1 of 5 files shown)


src/main/scala/co/ledger/wallet/daemon/database/core/WalletPoolDao.scala
```diff
@@ -67,6 +67,12 @@ class WalletPoolDao(poolName: String)(implicit val ec: ExecutionContext) extends
   override def findERC20OperationsByUids(a: Account, w: Wallet, filteredUids: Seq[ERC20OperationUid], offset: Int, limit: Int): Future[Seq[OperationView]] =
     erc20daoForWalletType(w.getWalletType: WalletType).findERC20OperationsByUids(a, w, filteredUids, offset, limit)
 
+  /**
+    * List erc20 operations from an account starting at specified bock height
```
mbasire-ledger, line 71: super bock height

src/main/scala/co/ledger/wallet/daemon/database/core/operations/EthereumDao.scala
```diff
@@ -37,6 +38,16 @@ class EthereumDao(protected val db: Database) extends CoinDao with ERC20Dao with
         "ORDER BY ercop.date " + order.value +
         s" OFFSET $offset LIMIT $limit"
 
+  private val erc20OperationFromBLockHeightQuery: (Int, String, Ordering.OperationOrder, Long, Int, Int) => SQLQuery =
```
mbasire-ledger, line 41: BLock => Block

src/main/scala/co/ledger/wallet/daemon/database/core/operations/EthereumDao.scala
```diff
@@ -37,6 +38,16 @@ class EthereumDao(protected val db: Database) extends CoinDao with ERC20Dao with
...
+    (accountIndex: Int, walletName: String, order: Ordering.OperationOrder, blockHeight: Long, offset: Int, limit: Int) =>
+      "SELECT ercop.uid as erc_uid, ercop.ethereum_operation_uid as eth_uid, ercop.receiver, ercop.value" +
+        "FROM wallets w, erc20_operations ercop, erc20_accounts ercacc, ethereum_accounts ethacc" +
+        s"WHERE w.name='$walletName' AND ethacc.idx='$accountIndex' " +
+        "AND ethacc.wallet_uid = w.uid AND ercop.account_uid = ercacc.uid AND ercacc.ethereum_account_uid = ethacc.uid " +
+        s"AND ercop.block_height >= '$blockHeight'" +
+        "ORDER BY ercop.date " + order.value +
+        s" OFFSET $offset LIMIT $limit"
+
```
mbasire-ledger, line 50: Change string concatenation by a `""" string """`  block on multi line.

src/main/scala/co/ledger/wallet/daemon/database/core/operations/EthereumDao.scala
```diff
@@ -150,47 +161,17 @@ class EthereumDao(protected val db: Database) extends CoinDao with ERC20Dao with
...
-            operationDate, BigInt(row.get[String]("status")).intValue()))
-        )
-      }
+      rowToFullOperationView(w, a.getIndex)
+    }
+  }
+
+  private def findErc20FullOperationViewFromBlockHeight(a: Account, w: Wallet, blockHeight: Long,
+                                                        offset: Int, limit: Int): Future[Seq[OperationView]] = {
+    logger.info(s"Retrieving erc20 operations for account : $a - limit=$limit offset=$offset from height ${blockHeight}")
```
mbasire-ledger, line 171: what information are serialized into the message about Account ? Are they all useful ?

src/main/scala/co/ledger/wallet/daemon/services/AccountsService.scala
```diff
@@ -291,9 +291,11 @@ class AccountsService @Inject()(daemonCache: DaemonCache, synchronizerManager: A
             publisher.publishOperation(view, account, wallet, accountInfo.poolName)
           ))
         val pushErc20Future: Future[Unit] = if (account.isInstanceOfEthereumLikeAccount) {
-          account.erc20Operations(wallet).map(_.map(view =>
-            publisher.publishERC20Operation(view, account, wallet, accountInfo.poolName)
-          ))
+          daemonCache.withAccountAndWalletAndPool(accountInfo) { (account, wallet, pool) =>
+            account.erc20OperationsFromHeight(pool, wallet, account, 0, Int.MaxValue, fromHeight.getOrElse(0)).map(_.map(view =>
```
mbasire-ledger, line 295: Could you add parameter name for values such as `0` or `Int.MaxValue` ?

mbasire-ledger (changes requested):

src/main/scala/co/ledger/wallet/daemon/database/core/operations/EthereumDao.scala
twilgenbus-ledger: I actually tried that idea! But I chose against it:
- because of the way it's done on the rest of this file (I feel I'd rather have the same bad convention everywhere, than many "good conventions")
- because it would mean I'd have to align my strings on the left AND I'd have to add `.replaceAll("\n", " ")` to prevent sending a multine SQL query (which is fine for execution, but might be impractical for logging purposes)

But I don't have a strong opinion on it, if you feel that those reasons aren't really constraining, I'd align myself with you!

src/main/scala/co/ledger/wallet/daemon/services/AccountsService.scala
twilgenbus-ledger: Good point! Will do

src/main/scala/co/ledger/wallet/daemon/database/core/operations/EthereumDao.scala
twilgenbus-ledger: After offline discssion, agree to go for `"""` but added `.replaceAll("\n", " ")`

src/main/scala/co/ledger/wallet/daemon/database/core/operations/EthereumDao.scala
twilgenbus-ledger: Good point, the log is currently useless. I've updated it.

mbasire-ledger (approved):

PR #347, opened 2021-03-18
Upgrade sbt to 1.4.8 and scala to 2.12.13

mbasire-ledger:
No need of the silencer plugin anymore thanks to the nowarn feature of scala 2.12.13


src/main/scala/co/ledger/wallet/daemon/clients/ScalaWebSocketClient.scala
```diff
@@ -86,10 +86,10 @@ class ScalaWebSocketClient extends co.ledger.core.WebSocketClient with Logging{
 
   }
 
-  // @silent is used here because we have no other way to walk around the
+  // @nowarn is used here because we have no other way to walk around the
   // ugly Java method without generating warning in Scala. @silent should
   // not be used as a no-brainer to by pass our fatal warning compiler option.
```
twilgenbus-ledger, line 91:
```suggestion
  // @nowarn is used here because we have no other way to walk around the
  // ugly Java method without generating warning in Scala. @nowarn should
  // not be used as a no-brainer to by pass our fatal warning compiler option.
```

twilgenbus-ledger (changes requested):
Blocked to wait for the new plugin to be released 👍

PR #349, opened 2021-03-19
Test tooling : catch sync end event

mbasire-ledger:


mbasire-ledger:
replaced by another PR

PR #360, opened 2021-04-01
add back curl for docker-compose healthcheck

CedricCabessa:
Since fb613fbb40e3, curl is missing but it is needed for `vault-integration`
docker-compose healthcheck:
https://github.com/LedgerHQ/vault-integration/blob/master/docker-compose.full.yml#L263

Without this, vault API CI cannot pass as we wait for all container to be
healthy before running tests

We need this patch in 3.0 and 3.1 branches


dud225 (approved):
Sorry I cleaned up the packages a bit too broadly.

twilgenbus-ledger (approved):
LGTM

sorry for the removal and thanks a lot for the fix

PR #361, opened 2021-04-06
BACK-1731 Cleaning PublisherModule dependencies

mbasire-ledger:
PublisherModule does not depend on DaemonCache anymore.
Pool goes from the account service to the AccountOperationsPublisher:
on publish, info are extracted from pool (see Publisher).

NOTE: Maybe payload info extraction should not be part of the Publisher's responsabilities.
NOTE2: it is a thin first step before testable PublisherModule components (next PR for BACK-1731)


twilgenbus-ledger (approved):
Noice!

PR #369, opened 2021-04-21
Add more buckets to serialEC hashmaps

gagbo:
Should fix collision issues (that trigger random latency spikes) on
auxiliary threads pools requested by libcore::WalletPool.

The collision is more likely to happen when multiple pools are ran on
the same WD, as each WalletPool creates 2 different
SerialExecutionContexts.

A collision means that threads on PoolA randomly have to wait on threads
running PoolB helpers before having some time allocated to them, thus
raising latency for seemingly no reason.

The number of buckets is raised to 1024, so that a collision only
happens if the last 10 bits of a hash function are common. BTC hopefully
showed us that it's rare enough


twilgenbus-ledger (approved):

drakoFukayu (approved):
LGTM

PR #371, opened 2021-04-26
AccountSynchronizer outputs sync time

twilgenbus-ledger:
Also removed bad process id with a random string


gagbo (approved):
Eventually we'll also have to find a way to register the enqueueing time for the accounts that overwhelm the maximum number of jobs, but the implementation is probably not as trivial

estrauser-ledger (approved):
Well done removing ids :+1:

docker/build.sh
```diff
@@ -4,6 +4,7 @@ set -euxo pipefail
 apt-get update
 apt-get install -qy apt-transport-https software-properties-common gnupg2
 apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 2EE0EA64E40A89B84B2DF73499E82A75642AC823
+add-apt-repository --remove "deb https://dl.bintray.com/sbt/debian /"
 add-apt-repository "deb https://repo.scala-sbt.org/scalasbt/debian /"
```
gagbo, line 8: Remove this, it’s no longer necessary

docker/build.sh
```diff
@@ -4,6 +4,7 @@ set -euxo pipefail
 apt-get update
 apt-get install -qy apt-transport-https software-properties-common gnupg2
 apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 2EE0EA64E40A89B84B2DF73499E82A75642AC823
+add-apt-repository --remove "deb https://dl.bintray.com/sbt/debian /"
```
gagbo, line 7: This should not work, it’s weird that it’s still there. (Bintray link will eventually stop working anyway)

PR #380, opened 2021-05-04
Add per-account resync from a given timestamp

gagbo:
Also adds the `since` query parameter to the DELETE account


RomanWlm (commented):
Great feature ! 

Some suggestions : 
- Does it make sens to add test to ensure that resync a single account does not impact an other ? 
- In production we often ends up with partially missing operation history, here we are testing when an account is compeltely empty as broken state. Do you think we should also cover partially broken cases ? Does it also work is there is only few operation missing in the middle of history ?

PR #386, opened 2021-05-17
Fix /operations/busy

gagbo:
Move /operations/xxxx routes to a higher route matching priority than
/operations/:uid

(2.10.3-rc.2)


estrauser-ledger (approved):

twilgenbus-ledger (approved):
LGTM!

PR #389, opened 2021-05-25
Bump libcore version

gagbo:
Maybe will fix the random end sync hangs ?


twilgenbus-ledger:
Closed to update to even latest version.

PR #399, opened 2021-06-11
Fix operations pagination

twilgenbus-ledger:


gagbo (approved):
Thanks for that, if you want to see what was missing tests-wise you can check the commit history on that branch.

We'll see later whether we merge it now or for a later version.

Happy vacation :)

